PR #103, opened 2015-01-04
Updated & proper logging.

### LINKED ISSUE #18: Add timestamps to console messages
This would be especially useful for tracebacks.

### LINKED ISSUE #72: Logging
Proper useful logging of all game mechanics (for debugging) is needed, including, but not limited to, **what** totem was given to whom and when.
Possibly held in memory until the end of a game and then written to disk.

Vgr255:
This change improves the old, obsolete, useless, awful and worthless logging feature by something that actually does a proper job at logging. This can be easily improved in the future. All instances of the old logger and the stdlib's logging module were removed. Timestamps can be displaying in both UTC +0000 and local <+/-> <offset> time. For example mine displays EST -0500. No gameplay changes, besides potential disk lag reduction.

***WARNING: This adds new USE_UTC and TIMESTAMP_FORMAT settings in the botconfig! You need to add them manually or the bot will crash.***

Edit: This will also close #18.


modules/wolfgame.py
```diff
@@ -4120,13 +3999,13 @@ def investigate(cli, nick, chan, rest):
...
                    " is a... \u0002{1}\u0002!").format(victim, vrole))
-    var.LOGGER.logBare(victim, "INVESTIGATED", nick)
+    debuglog("{0} ({1}) ID: {2} ({3})".format(nick, var.get_role(nick), victim, vrole))
     if random.random() < var.DETECTIVE_REVEALED_CHANCE:  # a 2/5 chance (should be changeable in settings)
         # The detective's identity is compromised!
         for badguy in var.list_players(var.WOLFCHAT_ROLES):
             pm(cli, badguy, ("Someone accidentally drops a paper. The paper reveals "+
                             "that \u0002{0}\u0002 is the detective!").format(nick))
-        var.LOGGER.logBare(nick, "PAPERDROP")
+        debuglog("{0} ({1}) PAPERDROP: {2}".format(nick, var.get_role(nick), " ".join(var.list_players(var.WOLFCHAT_ROLES))))
```
nyuszika7h: I'd probably say "PAPER DROP" here just to be consistent with things like "ELDER DEATH". Also, is listing the wolves really necessary here?

Vgr255: Listing wolves probably isn't necessary. I can do the PAPER DROP instead of PAPERDROP, I'll wait until it's reviewed a bit more before pushing that one though.

modules/wolfgame.py
```diff
@@ -5166,6 +5055,7 @@ def transition_night(cli):
         else:
             pm(cli, v_ghost, "You are a \u0002vengeful ghost\u0002.")
         pm(cli, v_ghost, who.capitalize() + ": " + ", ".join(pl))
+        debuglog("GHOST: {0} (Targets: {1}) Players: {2}".format(v_ghost, who.capitalize(), " ".join(pl)))
```
nyuszika7h: 1. Why is "targets" capitalized?
2. Not sure how useful listing the players is. I guess it can be useful but I don't know. Anyway, if they're listed, I think it should say `- players:` and the actual players should be comma-separated.

modules/wolfgame.py
```diff
@@ -6609,10 +6476,8 @@ def fpull(cli, nick, chan, rest):
         else:
             pm(cli, nick, "Process %s exited with %s %d" % (args, cause, ret))
 
-@cmd("fsend", admin_only=True, raw_nick=True, pm=True)
-def fsend(cli, raw_nick, chan, rest):
-    log_cmd(raw_nick, "fsend", "", rest)
-
+@cmd("fsend", admin_only=True, pm=True)
+def fsend(cli, nick, chan, rest):
```
nyuszika7h: Is this logged at all now?

modules/wolfgame.py
```diff
@@ -5853,7 +5731,7 @@ def allow_deny(cli, nick, chan, rest, mode):
                         del variable[acc]
                     msg = "\u0002{0}\u0002 (Account: {1}) is no longer {2} commands.".format(data[0], acc, "allowed any special" if mode == 'allow' else "denied any")
         elif var.ACCOUNTS_ONLY:
-            msg = "Error: \u0002{0}\u0002 is logged in to NickServ.".format(data[0])
+            msg = "Error: \u0002{0}\u0002 is not logged in to NickServ.".format(data[0])
```
nyuszika7h: You probably should push this directly to master rather than fixing it in this PR.

modules/wolfgame.py
Vgr255: 1. That's how it is for the message to ghost, but looking at it it does seem pointless. Removed.
2. Well I figured it couldn't hurt to list them, not sure. Modified it so it says something along the lines of: `GHOST: Vgr (target: wolves) - players: nyuszika7h, foo, bar`

modules/wolfgame.py
Vgr255: Yes. All admin commands are logged directly from the decorators.

modules/wolfgame.py
Vgr255: I'll add it in the commit description.

ilbelkyr: Please do fix this on master. It doesn't belong in this PR.

Vgr255: okay.jpg

nyuszika7h:
I updated your PR description to make sure GitHub will automatically close #18 upon merging this.

nyuszika7h:
There seem to be many things (fool win, lover death, traitor transformation, etc.) which don't appear to be logged at all now.

Vgr255:
Indeed. Going to be fixing this later today.

Issue #43, opened 2014-07-26
a new command needed: "!joinif <number>"


f:
Sometimes I want to play the game if at least 8 people is playing.

So it can be like that:

``` bash
fka: !joinif 8
lycan: You will be joined to the next 8 or more people game.
```

And, if anyone starts the game before 8 people are joined, it'll be started without me. But it'll remember my choice so whenever it's 8 people game I'll be joined. And after my name, I've to !joinif or !join again as usual.

IDK if I can tell well or not but I think you all got it :)

> Moved from https://github.com/skizzerz/lykos/issues/56

PR #113, opened 2015-02-15
Add the new ping-if feature.

Vgr255:
This adds a `!pingif` (and a few aliases, such as `!pingat` and `!pingme`) command to lykos. People who pick this option will see themselves added to the !ping list, regardless of their !away status, when the amount of players reaches or exceeds their preferences. `!pingif` can also be used to set how they want to be pinged (either right away when the sufficient amount of players have joined, added to the ping list but not pinged, or both). The bot will ping all players it can roughly 10 seconds after the last !join; this is done to reduce unnecessary who requests and pingspamming, which could cause both lag and annoyance. This is done through a separate thread, started when the first player joins, which exits when the game starts or the last player quits.

This is pretty much ready to merge, awaiting some feedback. I don't think there's anything else left to add or change, I think I've managed to please everyone with this.

Edit: This references #43 and sort of fills its purpose. It pings you instead of joining you, and does pretty much everything in a much saner way.

### DIFF
--- a/modules/wolfgame.py
+++ b/modules/wolfgame.py
@@ -333,7 +333,6 @@ def reset():
     var.ROLES = {"person" : []}
     var.JOINED_THIS_GAME = [] # keeps track of who already joined this game at least once (cloaks)
     var.JOINED_THIS_GAME_ACCS = [] # same, except accounts
-    var.PING_JOIN_TIMER = 0
     var.PINGED_ALREADY = []
     var.PINGED_ALREADY_ACCS = []
     var.NO_LYNCH = []
@@ -983,66 +982,67 @@ def toggle_altpinged_status(nick, value, old=None):
                         var.PING_IF_NUMS[old].remove(cloak)
 
 def join_timer_handler(cli):
-    while var.PHASE == "join":
-        with var.GRAVEYARD_LOCK:
-            if var.PING_JOIN_TIMER and var.PING_JOIN_TIMER + timedelta(seconds=10) <= datetime.now():
-                to_ping = []
-                pl = var.list_players()
+    var.PINGING_IFS = True
+    to_ping = []
+    pl = var.list_players()
 
-                var.PINGING_IFS = True
-                checker = []
-                chk_acc = []
+    checker = []
+    chk_acc = []
 
-                for num in var.PING_IF_NUMS_ACCS:
-                    if num <= len(pl):
-                        chk_acc.extend(var.PING_IF_NUMS_ACCS[num])
+    for num in var.PING_IF_NUMS_ACCS:
+        if num <= len(pl):
+            chk_acc.extend(var.PING_IF_NUMS_ACCS[num])
 
-                if not var.ACCOUNTS_ONLY:
-                    for num in var.PING_IF_NUMS:
-                        if num <= len(pl):
-                            checker.extend(var.PING_IF_NUMS[num])
+    if not var.ACCOUNTS_ONLY:
+        for num in var.PING_IF_NUMS:
+            if num <= len(pl):
+                checker.extend(var.PING_IF_NUMS[num])
 
-                for acc in chk_acc[:]:
-                    if acc in var.PINGED_ALREADY_ACCS:
-                        chk_acc.remove(acc)
+    for acc in chk_acc[:]:
+        if acc in var.PINGED_ALREADY_ACCS:
+            chk_acc.remove(acc)
 
-                for cloak in checker[:]:
-                    if cloak in var.PINGED_ALREADY:
-                        checker.remove(cloak)
+    for cloak in checker[:]:
+        if cloak in var.PINGED_ALREADY:
+            checker.remove(cloak)
 
-                if not chk_acc and not checker:
-                    var.PINGING_IFS = False
-                    var.PING_JOIN_TIMER = 0
-                    return
+    if not chk_acc and not checker:
+        var.PINGING_IFS = False
+        return
 
-                @hook("whospcrpl", hookid=387)
-                def ping_altpingers(cli, server, nick, ident, cloak, _, user, status, acc):
-                    if ('G' in status or is_user_stasised(user)[0] or not var.PINGING_IFS or
-                        user == botconfig.NICK or user in pl):
+    @hook("whospcrpl", hookid=387)
+    def ping_altpingers(cli, server, nick, ident, cloak, _, user, status, acc):
+        if ('G' in status or is_user_stasised(user)[0] or not var.PINGING_IFS or
+            user == botconfig.NICK or user in pl):
 
-                        return
+            return
 
-                    if acc and acc != "*":
-                        if acc in chk_acc and acc in var.PING_PREFS_ACCS and var.PING_PREFS_ACCS[acc] in ("once", "all"):
-                            to_ping.append(user)
-                            var.PINGED_ALREADY_ACCS.append(acc)
+        if acc and acc != "*":
+            if acc in chk_acc and acc in var.PING_PREFS_ACCS and var.PING_PREFS_ACCS[acc] in ("once", "all"):
+                to_ping.append(user)
+                var.PINGED_ALREADY_ACCS.append(acc)
 
-                    elif not var.ACCOUNTS_ONLY and cloak in checker and cloak in var.PING_PREFS and var.PING_PREFS[cloak] in ("once", "all"):
-                        to_ping.append(user)
-                        var.PINGED_ALREADY.append(cloak)
+        elif not var.ACCOUNTS_ONLY and cloak in checker and cloak in var.PING_PREFS and var.PING_PREFS[cloak] in ("once", "all"):
+            to_ping.append(user)
+            var.PINGED_ALREADY.append(cloak)
 
-                @hook("endofwho", hookid=387)
-                def fetch_altpingers(*stuff):
-                    var.PINGING_IFS = False
-                    var.PING_JOIN_TIMER = 0
-                    decorators.unhook(HOOKS, 387)
-                    if to_ping:
-                        to_ping.sort(key=lambda x: x.lower())
-                        cli.msg(botconfig.CHANNEL, "PING! {0} players! {1}".format(len(pl), " ".join(to_ping)))
+    @hook("endofwho", hookid=387)
+    def fetch_altpingers(*stuff):
+        # fun fact: if someone joined 10 seconds after someone else, the bot would break.
+        # effectively, the join would delete join_pinger from var.TIMERS and this function
+        # here would be reached before it was created again, thus erroring and crashing.
+        # this is one of the multiple reasons we need unit testing
+        # I was lucky to catch this in testing, as it requires precise timing
+        # it only failed if a join happened while this outer func had started
+        if 'join_pinger' in var.TIMERS:
+            del var.TIMERS['join_pinger']
+        var.PINGING_IFS = False
+        decorators.unhook(HOOKS, 387)
+        if to_ping:
+            to_ping.sort(key=lambda x: x.lower())
+            cli.msg(botconfig.CHANNEL, "PING! {0} players! {1}".format(len(pl), " ".join(to_ping)))
 
-                cli.who(botconfig.CHANNEL, "%nushaf")
-
-        time.sleep(5)
+    cli.who(botconfig.CHANNEL, "%nushaf")
 
 @cmd("join", "j", none=True, join=True)
 def join(cli, nick, chan, rest):
@@ -1096,12 +1096,8 @@ def join_player(cli, player, chan, who = None, forced = False):
         var.PHASE = "join"
         var.WAITED = 0
         var.GAME_ID = time.time()
-        var.PING_JOIN_TIMER = datetime.now()
         var.PINGED_ALREADY_ACCS = []
         var.PINGED_ALREADY = []
-        join_pinger = threading.Thread(None, join_timer_handler, args=(cli,))
-        join_pinger.daemon = True
-        join_pinger.start()
         if cloak:
             var.JOINED_THIS_GAME.append(cloak)
         if acc:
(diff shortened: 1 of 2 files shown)


ilbelkyr:
> This is also slow because it does /who on the channel while it could check var.USERS instead.

It *has* to do the `WHO`, same as the existing !ping command, to check for users being /away. If we don't do this here, we'd have to do it periodically anyway – and it'd be potentially outdated by the time we ping.

I'm not sure if any ircd rate-limits channel `WHO` very much; in chary at least it's a non-issue.

(There's the `away-notify` CAP but freenode doesn't support that yet; some other networks don't yet, either.)

Vgr255:
Oh yes, away status, I forgot about that one.

As discussed in -dev, I might add a new command to toggle between "get pinged as soon as a join makes the player count reach your preference" and not. I'm considering adding people to the !ping list regardless of this choice.

jacob1:
I would rather only be added to the ping list and not immediately pinged, so I agree that you should make that change. The immediate ping would be ok just to please everyone too :P.

modules/wolfgame.py
```diff
@@ -787,6 +787,99 @@ def fpinger(cli, nick, chan, rest):
     var.LAST_PING = None
     pinger(cli, nick, chan, rest)
 
+@cmd("pingif", "pingme", "pingat", "pingpref", pm=True)
+def altpinger(cli, nick, chan, rest):
+    """Pings you when the number of players reaches your preference."""
+    if not rest or (rest.isdigit() and int(rest) == 0):
+        if is_user_altpinged(nick)[0]:
+            msg = "Your ping preferences have been removed (Was {0}).".format(is_user_altpinged(nick)[1])
```
nyuszika7h: >Was

Why Is This Capitalized?

Vgr255: Looked better to me. Can make it lowercase if it's preferred.

ilbelkyr: Should be lowercase, yeah.

modules/wolfgame.py
```diff
@@ -895,6 +988,32 @@ def join_player(cli, player, chan, who = None, forced = False):
...
+            if acc in chk_acc:
+                to_ping.add(user)
+            elif not var.ACCOUNTS_ONLY and cloak in checker:
+                to_ping.add(user)
+
+        @hook("endofwho", hookid=387)
+        def actually_pingem(*stuff):
+            if not to_ping:
+                return
+            cli.msg(chan, "PING! {0} players! ".format(players) + " ".join(to_ping))
```
nyuszika7h: Are you seriously combining `.format()` and string concatenation? **ಠ_ಠ**

Vgr255: ....Whoops

modules/wolfgame.py
```diff
@@ -463,7 +466,7 @@ def pinger(cli, nick, chan, rest):
     var.PINGING = True
     TO_PING = []
 
-
+    pl = len(var.list_players())
```
skizzerz: This should be moved into on_whoreply to account for cases where someone does !ping and someone else joins before the WHO reply comes in -- that way the number of players can always be accurate.

Also, don't use pl as the var name, normally that name stands for "player list" aka the actual list of players, not the length of the player list; pick a different name so we aren't overloading the semantic meaning of variable names.

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
...
+            toggle_altpinged_status(nick, rest)
+    else:
+        cli.notice(nick, "You need to enter a non-negative integer.")
+
+def is_user_altpinged(nick):
+    if nick in var.USERS.keys():
+        cloak = var.USERS[nick]["cloak"]
+        acc = var.USERS[nick]["account"]
+    else:
+        return False, None
```
skizzerz: Style nit: If returning a tuple, please surround it in parenthesis. This makes it obvious to a casual observer that a) you are returning a tuple, and b) that returning a tuple was intentional as opposed to whatever else commas let you do

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
     var.LAST_PING = None
     pinger(cli, nick, chan, rest)
 
+@cmd("pingif", "pingme", "pingat", pm=True)
+def altpinger(cli, nick, chan, rest):
+    """Pings you when the number of players reaches your preference."""
+    if not rest or (rest.isdigit() and int(rest) == 0):
```
skizzerz: Not entirely sure that having the param blank should be considered the same as clearing it. It may be more useful for an empty !pingif to simply tell you what your current preference is.

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
...
+        if is_user_altpinged(nick)[0]:
```
skizzerz: I'd probably save is_user_altpinged(nick) off to a variable -- or even better, decompose it into two variables, e.g. (is_altpinged, altping_amount) = is_user_altpinged(nick). Then you can use is_altpinged and altping_amount later in the function to give more semantic meaning to all of your random checks as opposed to indexing into some arbitrary tuple.

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
...
+            else:
+                cli.notice(nick, "You need to enter a number.")
+        else:
+            if chan == nick:
+                pm(cli, nick, "You do not have any preferences set.")
+            else:
+                cli.notice(nick, "You do not have any preferences set.")
+    elif rest.isdigit():
+        rest = int(rest)
+        if rest >= var.MAX_PLAYERS:
```
skizzerz: var.MAX_PLAYERS changes based on gamemode (or can change). If you absolutely MUST limit it, I'd rather go for some arbitrarily high number like 9999. There is basically no downside to picking a number higher than how many people would feasibly join as you'd just never get pinged.

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
...
+        return False, None
+    if acc and acc != "*":
+        if acc in var.PING_IF_PREFS_ACCS.keys():
+            return True, var.PING_IF_PREFS_ACCS[acc]
+    elif not var.ACCOUNTS_ONLY and cloak in var.PING_IF_PREFS.keys():
+        return True, var.PING_IF_PREFS[cloak]
+    return False, None
+
+def toggle_altpinged_status(nick, value, old=None):
+    # nick should be in var.USERS; if not, let the error propagate
```
skizzerz: nick won't be in var.USERS if it is a fake nick. Speaking of which, how are you handling those? You should probably have !pingif silently reject fake nicks instead of blowing up.

modules/wolfgame.py
```diff
@@ -787,6 +795,189 @@ def fpinger(cli, nick, chan, rest):
...
+            var.set_ping_pref(cloak, "all")
+
+    if chan == nick:
+        pm(cli, nick, msg.format(botconfig.CMD_CHAR))
+    else:
+        cli.notice(nick, msg.format(botconfig.CMD_CHAR))
+
+def join_timer_handler(cli):
+    while var.PHASE == "join":
+        with var.GRAVEYARD_LOCK:
```
skizzerz: This is a different thread than the reaper, it should probably use a different lock. Probably doesn't really matter, but semantics!

Vgr255: Well, it doesn't matter at all, since both threads that use this are mutually exclusive; one runs when a game is not running, the other runs when a game is running. It seemed pointless to create a new lock that is only used for that, rather than using one which is already there and unused at the time we need to use it.

modules/wolfgame.py
```diff
@@ -787,6 +796,203 @@ def fpinger(cli, nick, chan, rest):
...
+                        if num <= len(pl):
+                            checker.extend(var.PING_IF_NUMS[num])
+
+                    for num in var.PING_IF_NUMS_ACCS:
+                        if num <= len(pl):
+                            chk_acc.extend(var.PING_IF_NUMS_ACCS[num])
+
+                    @hook("whospcrpl", hookid=387)
+                    def ping_altpingers(cli, server, nick, ident, cloak, _, user, status, acc):
+                        if ('G' in status or '+' in status or is_user_stasised(user)[0] or
```
brrr2: Shouldn't there be a check here to see if the user has already joined the game?

modules/wolfgame.py
```diff
@@ -471,11 +476,17 @@ def on_whoreply(cli, server, dunno, chan, dunno1,
         if not var.PINGING: return
         if user in (botconfig.NICK, nick): return  # Don't ping self.
 
-        if ('G' not in status and '+' not in status and not
-            is_user_stasised(user)[0] and not is_user_away(user)):
-
-            TO_PING.append(user)
-
+        if ('G' not in status and '+' not in status and not is_user_stasised(user)[0]):
```
skizzerz: Per brrr's comment, this should check to see if the user actually joined, not if the user has voice, for two reasons:
1) There are cases where people may be voiced but are not playing, not every channel is ##werewolf after all and we should not impose that a channel MUST be run a particular way when voicing users in order to allow the bot to work.
2) Voice is not necessarily indicated by the + character, so if you wanted to check this for real you'd actually need to hit up ISUPPORT to check which character corresponds to +v.

modules/wolfgame.py
```diff
@@ -787,6 +798,249 @@ def fpinger(cli, nick, chan, rest):
...
+        else:
+            msg.append("You do not have any preferences set.")
+
+    elif rest[0].isdigit() or (len(rest) > 1 and rest[1].isdigit()):
+        if rest[0].isdigit():
+            num = int(rest[0])
+        else:
+            num = int(rest[1])
+        if num > 999:
+            msg.append("Please select a number between 0 and {0}.".format(var.MAX_PLAYERS))
```
skizzerz: update message to reflect the limit is 999, not var.MAX_PLAYERS

modules/wolfgame.py
```diff
@@ -787,6 +798,249 @@ def fpinger(cli, nick, chan, rest):
...
+            msg.append("Your ping preferences are already set to {0}.".format(num))
+        elif altpinged:
+            msg.append("Your ping preferences have been changed from {0} to {1}.".format(players, num))
+            toggle_altpinged_status(nick, num, players)
+        else:
+            msg.append("Your ping preferences have been set to {0}.".format(num))
+            toggle_altpinged_status(nick, num)
+
+    if (not acc or acc == "*") and var.ACCOUNTS_ONLY:
+        pass # we've taken care of this above already
```
skizzerz: This is really wonky. Instead of adding a dummy check here, just make the first ACCOUNTS_ONLY check actually perform the pm/notice and then return from the function. Otherwise I can see this quietly biting us later if we add some additional feature to pingif without also adding yet another dummy check for this

modules/wolfgame.py
jacob1: 999 is a weird number, instead of printing 999 in the message, maybe should instead just say "That number was too large" (could even raise the limit to 2**31 like stasis, dunno)

modules/wolfgame.py
```diff
@@ -787,6 +798,249 @@ def fpinger(cli, nick, chan, rest):
...
+                    var.PING_PREFS_ACCS[acc] = "all"
+                    var.set_ping_pref_acc(acc, "all")
+            elif cloak in var.PING_PREFS.keys() and var.PING_PREFS[cloak] == "all":
+                msg.append("You are already set to be added to the {0}ping list as well as being pinged immediately when enough players have joined.")
+            else:
+                msg.append("You will now be added to the {0}ping list as well as being pinged immediately when your preferred amount of players is reached.")
+                var.PING_PREFS[cloak] = "all"
+                var.set_ping_pref(cloak, "all")
+        else:
+            msg.append("Invalid parameter. Please enter a non-negative integer or a valid preference.")
```
skizzerz: Maybe silently fail here? That way I can do something like "!pingif 6 because I hate small games" -- a lot of our other commands allow that kind of construct where the user can insert arbitrary text after the parameters as kind of an explanatory note; I see no reason to prevent it here as well (especially since the command actually succeeds at setting your pref to 6 in addition to giving you the error message)

modules/wolfgame.py
```diff
@@ -787,6 +798,249 @@ def fpinger(cli, nick, chan, rest):
...
+                    var.PINGING_IFS = False
+                    var.PING_JOIN_TIMER = 0
+                    decorators.unhook(HOOKS, 387)
+                    if to_ping:
+                        to_ping.sort(key=lambda x: x.lower())
+                        cli.msg(botconfig.CHANNEL, "PING! {0} players! {1}".format(len(pl), " ".join(to_ping)))
+
+                cli.who(botconfig.CHANNEL, "%nushaf")
+
+        time.sleep(5)
```
skizzerz: This is still horribly done. A better construction would be instead of having a dedicated thread for this, to set a 10-second timer on each !join that calls the handler function. Each subsequent join would then cancel the current timer and set a new one for 10 seconds, to ensure that the ping always happens 10 seconds after the latest join.

modules/wolfgame.py
Vgr255: The message lying about the maximum amount is intentional. In normal ways you wouldn't need to exceed the maximum amount of allowed players. I could change this to state that the number is too large.

modules/wolfgame.py
Vgr255: Yeah, definitely is. I don't really know what you mean though, and I don't know how I could make it better. I'm not good at threading and you know it.

modules/wolfgame.py
Vgr255: I changed so it only warns you if nothing else suceeded, Should be fine.

modules/wolfgame.py
nyuszika7h: What's the problem with using a separate thread? Just use `threading.Timer`.

Vgr255:
So this PR is ready to be merged, I squashed every bug I could, and all options that were requested were added. Awaiting further feedback.

PR #119, opened 2015-03-08
Allow amnesiac to be the default role.

Vgr255:
This adds support for any arbitrary role to be the default role.

### DIFF
--- a/modules/wolfgame.py
+++ b/modules/wolfgame.py
@@ -1476,9 +1476,9 @@ def stats(cli, nick, chan, rest):
         if amnrole in ("village elder", "time lord"):
             amnrole = "villager"
         elif amnrole == "vengeful ghost":
-            amnrole = var.DEFAULT_ROLE
+            amnrole = var.get_default_role()
         elif amnrole == "traitor" and var.HIDDEN_TRAITOR:
-            amnrole = var.DEFAULT_ROLE
+            amnrole = var.get_default_role()
         if amnrole != "amnesiac":
             amn_roles["amnesiac"] += 1
             if amnrole in amn_roles:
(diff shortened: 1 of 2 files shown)


skizzerz:
There are a number of issues with this, but I will not be back until Tuesday to fully address these (very limited internet in the meantime). Do not merge this.

skizzerz:
Don't hardcode amnesiac to villager if it is the default role, add a new variable that allows for selecting either villager or cultist as what the amnesiac starts off as. Stats and winner determination should be based on this new variable as well if amnesiac is the default role. If you do this, you're also only a step off from allowing arbitrary default roles in that you'd check team affiliation for win conditions/stats. So you can do something like
```python
default_team = "wolf" if (var.AMNESIAC_DEFAULT if var.DEFAULT_ROLE == "amnesiac" else var.DEFAULT_ROLE) in var.WOLFTEAM_ROLES else "villager"

if default_team == "villager":
    count += len(var.ROLES["village elder"] + var.ROLES["time lord"] + var.ROLES["vengeful ghost"])
    count -= len([p for p in var.CURED_LYCANS if p in var.ROLES["villager"]])
    count += bitten_roles["village elder"] if "village elder" in bitten_roles else 0

# etc
```
(yes that sets the default team to villagers if a neutral is selected as the default. That is intentional because things would be very complicated and confusing if that were not the case)

skizzerz:
Other random thoughts:
1) On night 1 (assuming that var.AMNESIAC_NIGHTS > 1), make sure you tell amnesiac that they are a whatever var.AMNESIAC_ROLE is set to instead of hardcoding it to villager.
2) If var.AMNESIAC_NIGHTS == 1, then it may be better to skip the "your amnesia clears and you are a xxxxxx" and just tell them the actual role.
3) var.AMNESIAC_DEFAULT should be restricted to either "villager" or "cultist" (just like how var.DEFAULT_ROLE used to be). Games should refuse to start if this is not the case.
4) If made to accept arbitrary roles, var.DEFAULT_ROLE should not be a template. If it is the game should refuse to start.

skizzerz:
Closing this PR, feel free to reopen when it is fixed up and ready to look at again.

PR #125, opened 2015-04-06
Add the new logger

Vgr255:
This adds the new logger module, which I have spent about 2 months developing now. This improves the handling of the logging we currently have, is less error-prone than the small one I made some time ago which we currently use, and brings in some interesting functionality that we may be able to implement in the future.

This is a large and possibly confusing module, however it is all ready-to-use with the setup I made, with very few changes to the code, aside from changing the few functions calls to lambda functions to convert the current calls into ones the logger can interpret.

This is ready to merge, and has been tested for all locations where the logger is used. Due to the nature of this module, it is extremely unlikely that it broke something else.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -3,6 +3,7 @@ import datetime
 import time
 
 import botconfig
+from src import _logger
 from src import settings as var
 
 # Todo: Allow game modes to be set via config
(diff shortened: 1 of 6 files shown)


nyuszika7h:
As I and @jacob1 have said, I don't see any advantage, but it probably doesn't hurt either. Not really sure if we want to include a new logger if the current one works fine.

skizzerz:
I also echo @nyuszika7h's concerns here, this seems to be a solution looking for a problem. What, exactly, is wrong or deficient with our current logger? If the answer to that is "nothing" then this PR is wholly pointless. If there is something, a better course of action would be to identify the areas of improvement in our current logger and apply incremental improvements to it. Very rarely is ripping out all of the old code and putting in brand new (and in the case of this PR, bloated with "features" we would never use) merited, and I am not convinced that is what is needed here.

@Vgr255: Please identify what is wrong/deficient with the current logger in the comments here, answer how this PR solves those problems, and answer why this new logger is needed in order to solve the problem as opposed to an improvement on the current logger.

Vgr255:
As it is right now, the current logger works around some of the problems (instead of working towards resolving them), and it "just works" (which, really, could be said for most of lykos). I wanted to rewrite it so that it works properly, and ended up doing it as a separate project instead. The original logger suffered from my (lack of) programming knowledge at the time, which I obviously improved. The goal of this PR is not fixing the logger per se (since it does work as intended), but more of an improvement towards lykos in general, by making it OO as you wanted to make it for a while. Granted, there are features that lykos wouldn't make use of, and yes, improving the current logger could also be fine, but if we're going to work towards of goal of making lykos overall better, this is one of many steps towards that goal. I can see the relunctance of getting a whole module out of nowhere into lykos, and I understand that you'd rather see the current one improve. However, I believe that between this module and the OOO, lies only a thin line which I'm perfectly fine of crossing if this is going to bring in an improvement to the code. I believe this is one of such improvements, and yet anothet step towards making lykos entirely object-oriented.

skizzerz:
Can you please identify the aforementioned problems that were "worked around"? Without sharing what problems this PR hopes to solve, I cannot begin to review it on its merits because I have no idea what I'd be looking for. With any reasonably-sized codebase, it is impossible for one person to be intimately familiar with all of it. Since you are the most familiar with the logger due to having written the thing, sharing the proper context assists others in reviewing modifications to it and offering more useful suggestions if needed. Or, to put it a different way, I'm not saying there aren't issues with the current logger, but if there are I am completely unaware of them. I feel others may be in similar situations. So please, inform us with specifics on what this PR solves instead of a general "it's better, trust me" statement.

Vgr255:
I wanted to address these in my comment, but I'm in a rush tonight. Will point these when I have time, either later tonight (unlikely) or tomorrow.

Vgr255:
I know there was something about files not being properly written to (that was fixed before I committed though), but it's pretty far and I don't remember what it had problems with. There was also the crash with ಠ\_ಠ characters, which couldn't have been fixed any other way anyway so that doesn't count (both loggers handle these properly). As it stands right now, the logger doesn't have any issues other than its pretty hackish code, which, as I stated, is the main point behind this PR. I am not saying that we __need__ this new logger, but I'm saying that it's one of many steps towards an overall more object-oriented code base for lykos. Again, it's a code improvement, not a bugfix.

Vgr255:
As I see it, it's probably not getting merged, and I'm in the process of reworking it entirely (I wasn't happy about it), so I might as well close it and maybe submit another one 6 months from now. No need to keep that open; the branch is also deleted.

PR #131, opened 2015-05-08
Add event system

skizzerz:
Also fixes and reenables evilvillage (as an example of how the events system can be used).


ilbelkyr:
At a glance, looks quite nice. I assume this has been tested.

There's definitely a lot of potential uses for that feature, and fixing evilvillage is a neat demonstration of that. +1 from me.

PR #132, opened 2015-05-10
Rename src to core

skizzerz:
Name makes more sense, since it's the bot core. Eventual dichotomy for bot operators is files in / (with exception of wolfbot.py) are meant to contain local customizations (config in botconfig.py, game modes in gamemodes.py, etc.), and stuff in core isn't meant to be edited by bot operators.


nyuszika7h:
Why `core`? I think `src` makes perfect sense.

skizzerz:
src is a standin from C, where you have a src with all of the source files (.c) and include with all of the headers. Core, on the other hand, indicates that the code therein forms the "core" code of the bot, giving an indication to operators that they shouldn't be editing it if they want support or easy updates. I really have no preference either way (I'm pretty sure I was the one that suggested src to begin with), it's just do we want the name to hold a semantic meaning or not.

The main important part is that everything is in that folder, as new refactoring will add more files to the root folder that aren't tracked in git and are meant for bot operators to customize their bot for their own channels (adding new roles, gamemodes, etc.), and this PR doesn't change that at all.

nyuszika7h:
I don't think this would help operators who don't understand that in the first place. It would be better to add something to the README instead.

PR #133, opened 2015-05-14
Add "kpon" as join trigger

MJ94:
Because I (and others) frequently typo "join" as "kpon".


jacob1:
This command name doesn't make any sense, and we aren't going to start adding typos as command names. If it is a constant issue, look into adding some kind of autocorrect in your client, I know hexchat supports auto replace at least.

PR #135, opened 2015-05-20
Notify users who have +g set in the channel

ilbelkyr:
There's a notification when users PM you while you're +g, but it isn't
always noticed - so highlight them in the channel as well.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6187,6 +6187,13 @@ def start(cli, nick, chan, forced = False, restart = ""):
 
     if restart:
         var.PHASE = None # allow transition_* to run properly if game was restarted on first night
+
+    var.UMODE_G = set()
+
+    @hook("716", hookid=716) # oyoyo doesn't know 716 / targumodeg
+    def umode_g(cli, prefix, me, target, *rest):
+        var.UMODE_G.add(target)
+
     var.FIRST_NIGHT = True
     if not var.START_WITH_DAY:
         var.GAMEPHASE = "night"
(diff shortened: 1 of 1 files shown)


nyuszika7h:
Perhaps change it to "unset it or do `/accept $botnick`"? But otherwise, looks good.

nyuszika7h:
Since even villagers are PMed now, this shouldn't reveal the existence of a special role, right?

skizzerz:
Maybe should check for CALLERID=g in ISUPPORT before enabling these hooks, but looking at other ircds out there none of them overload this numeric to mean something else so it should be fine as-is.

I echo @nyuszika7h's comment on adding `/accept <botnick>` to the sent message so they can allow lykos while still keeping +g on.

Finally, this code should really be in a different file (src/handler.py seems like a good candidate) so that the code is actually together rather than being intermixed with gameplay code. In the gameplay code itself you can add some events for `on_start` and `after_role_pms` then have the code in handler.py listen in on those events and insert the necessary IRC hooks.

ishanyx:
You need to do !myrole anyway to see what you missed, and if you do that in PM it'll accept lykos automatically.

nyuszika7h:
(Label requested by @Vgr255. If you don't like it, blame him.)

Vgr255:
ಠ_ಠ

nyuszika7h:
*pokes @ilbelkyr*

nyuszika7h:
Is this ever going to be done? :P

skizzerz:
Closing for now, reopen when ready.

nyuszika7h:
There might be another problem. If someone sets `+g` after the first night PM, lykos might reveal that they are a special role, as villagers are only PMed on night 1.

jacob1:
This only works night one anyway, and should probably stay that way for that reason. If they set +g after already receiving a PM, and then never do /accept, that is their fault ...

PR #136, opened 2015-05-20
!pingif advertisement

ilbelkyr:
This prepares for the planned deprecation of opt-out `!ping` in favor of `!pingif`. Additionally, to make up for the removal of using `!ping` to add 30 seconds of wait time, this adds a token bucket that `!wait` uses.


nyuszika7h:
Looks good to me.

skizzerz:
LGTM. Can merge now, but whomever merges should also reset the away list in the db before pushing it to lykos.

PR #143, opened 2015-05-28
Add the arsonist role

Vgr255:
Mostly tested, should work fine, but since this touches some crucial elements (transition_day), this should be reviewed before landing in master.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -1301,6 +1301,9 @@ def stats(cli, nick, chan, rest):
         if role in amn_roles:
             count += amn_roles[role]
 
+        if role in var.FAKE_ROLES:
+            count += var.FAKE_ROLES[role]
+
         if role == rs[0]:
             if count == 1:
                 vb = "is"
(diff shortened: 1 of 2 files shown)


src/wolfgame.py
```diff
@@ -3279,10 +3293,89 @@ def transition_day(cli, gameid=0):
         message.append("A chilling howl was heard last night. It appears there is another werewolf in our midst!")
         novictmsg = False
 
+    varkills = []
```
skizzerz: Unnecessary, simply use bywolves instead. bywolves should **never** be removed from, so it can be used to see if wolves targeted someone or not. onlybywolves is used if you want to ensure that someone was ONLY targeted by wolves and nobody else.

src/wolfgame.py
```diff
@@ -5654,6 +5795,25 @@ def transition_night(cli):
...
+        random.shuffle(pl)
+        pl.remove(ars)
+        for house in var.BURNED_HOUSES:
+            if house in pl:
+                pl.remove(house)
+        if ars in var.PLAYERS and not is_user_simple(ars):
+            pm(cli, ars, ("You are an \u0002arsonist\u0002. You may throw a molotov " +
+                          "at someone's house, killing who's inside, if there are any. " +
+                          'Use "throw <nick>" to throw a molotov at a player, or "pass" ' +
+                          'to not throw any molotov tonight. You get \u0002{0}\u0002 molotovs.').format(
```
skizzerz: This message sucks.

skizzerz:
Per discussion in -dev, I don't think this role is quite ready. I have a zany idea for making it a bunch more interactive while still keeping the same general theme (e.g. burning down houses where inhabitant's roles are hidden should they die from it), but will require a bunch more work. Won't write full notes down in github until we a) decide if we want to do that, and b) hammer out some of the finer details.

src/wolfgame.py
Vgr255: HELP MY LITERARY SKILLS SUCK USE burn <nick> YOU HAVE {0} SHOTS

Vgr255:
With neither @skizzerz and I working on it lately, I think it's safe to say this PR can be closed at least for now. The branch can be kept around, but we'll need to add the whole new gameplay mechanic later on.

PR #145, opened 2015-06-01
Change decorators

Vgr255:
Old decorators were silly and stupid, and we certainly did not need a generator for these. Moreover, it was a bunch of intrinsic namespace mangling, which we try to avoid as much as possible. I change those to be classes, so this means they are more customizable and less namespace-mangled-restricted-whatever. This works fine, waiting on other people's suggestions.

### DIFF
--- a/src/decorators.py
+++ b/src/decorators.py
@@ -1,167 +1,222 @@
-# Copyright (c) 2011, Jimmy Cao
-# All rights reserved.
-
-# Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
-
-# Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
-# Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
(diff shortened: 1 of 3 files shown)


src/decorators.py
```diff
@@ -1,167 +1,222 @@
...
-
-# Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
-# Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
-# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# Old, obsolete & original code by jcao219
+# rewritten by Vgr
```
skizzerz: I'd just remove these comments entirely. We have a LICENSE file to show what license the codebase is under, don't need to have attribution in the individual source files as well (git does attribution for who pretty much wrote what file anyway if you want to know that versus simply what license the codebase is under)

src/decorators.py
```diff
@@ -1,167 +1,220 @@
...
-                    if not allowed:
-                        return
-                if nick in var.USERS.keys() and var.USERS[nick]["account"] != "*":
-                    acc = var.USERS[nick]["account"]
+COMMANDS = defaultdict(list)
+HOOKS = defaultdict(list)
+
+class cmd:
+    def __init__(self, *cmds, raw_nick=False, admin_only=False, owner_only=False,
+                 chan=True, pm=False, join=False, none=False, game=False, playing=False, roles=()):
```
skizzerz: Consider condensing the join, none, game params into a single phase param, takes in a tuple/list/whatever like phase=("join","day","night") (could have "game" as an alias for both "day" and "night")

(I believe @nyuszika7h is the one that originally came up with this idea, just stating it in github comments for posterity's sake)

Vgr255: As I stated in -dev, it's about the messages. If we can all agree to "no error message" then I'll change them.

jacob1:
commenting on this pull request because Vgr told me to

Vgr255:
This should be pretty much ready to merge now. The `rate_limit` decorator was suggested by @nyuszika7h and I may add it if it's really necessary, otherwise this is ready.

PR #146, opened 2015-06-04
New roles, protection semantics, and alpha fixes

skizzerz:
This commit adds 3 new roles and changes the semantics of
kills/protections to work more like the impatience and pacifism totems.

- Add mystic and wolf mystic roles which can divine the number of
  players still alive on the opposing team each night. These roles are
  meant for games where role reveal is off as !stats gives the same info
  otherwise.
- Add fallen angel which is a corrupted version of the guardian angel,
  created when a guardian angel is bitten by alpha wolf. This role is
  incredibly potent as they bypass all forms of protection on their
  target. GAs corrupted this way also automatically gain the assassin
  template.
- Protections/kills now stack and cancel each other out. As such, 2
  kills but only 1 protection will result in the kill still happening.
- Fix help text for active alpha wolf to show the reworked bite command.
- If alpha bites someone who is going to die via other means that night,
  refund the bite.
- If the bite was unsuccessful due to whatever reason and was refunded,
  give a different message to alpha wolf to let them know that.
- Make adding more wolf roles in the future easier to do by replacing
  some hardcoded lists with var.WOLF_ROLES.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -1374,8 +1374,8 @@ def hurry_up(cli, gameid, change):
                 voters.append(v)
         for v in voters:
             weight = 1
-            imp_count = sum(1 if p == v else 0 for p in var.IMPATIENT)
-            pac_count = sum(1 if p == v else 0 for p in var.PACIFISTS)
+            imp_count = var.IMPATIENT.count(v)
+            pac_count = var.PACIFISTS.count(v)
             if pac_count > imp_count:
                 weight = 0 # more pacifists than impatience totems
             elif imp_count == pac_count and v not in var.VOTES[votee]:
@@ -1455,8 +1455,8 @@ def chk_decision(cli, force = ""):
             if v in pl and v not in voters and v != votee and v not in var.WOUNDED and v not in var.ASLEEP:
                 # don't add them in if they have the same number or more of pacifism totems
                 # this matters for desperation totem on the votee
-                imp_count = sum(1 if p == v else 0 for p in var.IMPATIENT)
-                pac_count = sum(1 if p == v else 0 for p in var.PACIFISTS)
+                imp_count = var.IMPATIENT.count(v)
+                pac_count = var.PACIFISTS.count(v)
                 if pac_count >= imp_count:
                     continue
 
@@ -1466,8 +1466,8 @@ def chk_decision(cli, force = ""):
                 impatient_voters.append(v)
         for v in voters[:]:
             weight = 1
-            imp_count = sum(1 if p == v else 0 for p in var.IMPATIENT)
-            pac_count = sum(1 if p == v else 0 for p in var.PACIFISTS)
+            imp_count = var.IMPATIENT.count(v)
+            pac_count = var.PACIFISTS.count(v)
             if pac_count > imp_count:
                 weight = 0 # more pacifists than impatience totems
             elif imp_count == pac_count and v not in var.VOTES[votee]:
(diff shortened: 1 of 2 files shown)


jacob1:
For wolf mystic, it is notified of some roles that it probably doesn't need to be, like vengeful ghost or lycan (these aren't necessarily against them). And maybe time lord too? All the rest of the neutrals are in TRUE_NEUTRAL_ROLES.

Also i'm not sure why fallen angel being alive should mean wolves can kill monster. Just seems like a random thing you would have to memorize, and it would only affect 'random' games.

There is also some comment in there about how if we add the mystic stuff to !myrole it would need to remember the old message. But this just gets outdated so quickly that i'm not sure if we should put it in !myrole at all. Someone else can try if they really want.

Besides that, it looks awesome :). I'll just assume all the transition_day changes work (you did test them), because that is a lot to confirm works exactly the same >_>

jacob1:
Oh, I think you also said something in channel about adding a separate message for bitten that says "you will turn in X days", and then turning off the really long one for people with !simple on. I would like this, even if you don't disable the long message, just adding a separate message explaining what that means would be nice.

skizzerz:
Fallen Angel bypasses any and all protections against being killed, which includes the innate protection monster has. Making it *not* work for monster is the exception, imo. I was considering making them able to kill visiting harlots as well (either they linger around until they come home or track them to where they wind up and kill both people there as if they targeted the victim), but decided to hold off on that for right now because it would be complicated to implement.

As for wolf mystic, probably does need some modification; at the very least it should discount hidden roles like amnesiac, vengeful ghost, time lord, and village elder. The point of the count, however, isn't a "this is how many people are against you" but rather "this is how many non-wolfteam has special powers" -- I feel lycan at the very least should count as part of that.

jacob1:
I don't think monster not dying is due to any type of "protection", but if you think fallen angel should ignore that protection too then ok, sure.

Also your description for mystics said "can divine the number of players still alive on the opposing team each night". Lycan technically is on the opposing team so sure. But hidden roles should be removed from the list yeah :P

skizzerz:
Lore-wise, the monster draws its power from the darkness, so at night when it detects wolves attacking it, it can quickly fade into the darkness and escape (leaving nothing left for wolves to nom). Hunter can shoot it from afar meaning it doesn't have a chance to react, and death totem takes effect at dawn where the monster's power is insufficient to stop it. The vengeful ghost and fallen angel can follow it through the darkness and kill it as well.

src/wolfgame.py
```diff
@@ -3879,6 +4011,9 @@ def check_exchange(cli, actor, nick):
             if nick_role == "shaman":
                 pm(cli, actor, "You have a \u0002{0}\u0002 totem.".format(nick_totem))
             var.TOTEMS[actor] = nick_totem
+        elif nick_role == "mystic":
+            numevil = len(var.list_players(var.WOLFTEAM_ROLES))
+            pm(cli, actor, "There are \u0002{0}\u0002 evil villager{1} still alive.".format(numevil, "s" if numevil != 1 else ""))
```
Michagogo: Erm, seems like that will say "There are 1 evil villager still alive." Should be "is".

src/wolfgame.py
```diff
@@ -3895,15 +4030,17 @@ def check_exchange(cli, actor, nick):
                     pl[i] = player + " (cursed)"
 
             pm(cli, actor, "Players: " + ", ".join(pl))
-            angry_alpha = ''
+            if actor_role == "wolf mystic":
+                # # of special villagers = # of players - # of villagers - # of wolves - # of neutrals
+                numvills = len(ps) - len(var.list_players(var.WOLFTEAM_ROLES)) - len(var.list_players(("villager", "vengeful ghost", "time lord", "amnesiac", "village elder", "lycan"))) - len(var.list_players(var.TRUE_NEUTRAL_ROLES))
+                pm(cli, actor, "There are \u0002{0}\u0002 special villager{1} still alive.".format(numvills, "s" if numvills != 1 else ""))
```
Michagogo: Same here -- should only say "are" if numvills != 1.

PR #149, opened 2015-07-01
Mass refactor of the data types

Vgr255:
This changes a lot of the internal workings by replacing most of the lists by sets where possible. OrderedDict is used for ROLE_GUIDE, list_players now uses join order, and role_order was superseded by the ROLE_GUIDE ordering. Should be ready to merge. Needs reviewing.

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -2,7 +2,7 @@ import fnmatch
 import math
 import random
 import sqlite3
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 
 import botconfig
 from src import events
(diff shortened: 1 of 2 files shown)


Vgr255:
This has been tested, and is pretty much ready to merge, save for one obscure bug which makes literally no sense; at some random point (and this happened out of nowhere, all of a sudden) wolves cannot kill anyone - the bot doesn't reply, and var.KILLS isn't modified. Other than that, I don't know why commits from master landed in here. Poke me when the PR is ready to merge, I'll squash everything in one commit.

jacob1:
I was testing it and I found a few bugs (some major):
it errors every time village wins. This is because it checks cubs == 0 but the cubs variable isn't defined (removed for some reason). Can be fixed with `cubs = len(var.ROLES.get("wolf cub", ()))`

It also errored when I did !rforce person quit, not sure if that is specifically from this pull request (I did this before though), it can be fixed with `tgt = copy.copy(var.ROLES[who])` instead of just `tgt = var.ROLES[who]`

Desperation totem also doesn't work at all, the main issue is that the list of people who have voted someone is a set and not a list, and it needs to do target = voters[-1] to get the last person who voted.

I also noticed why kills weren't working in that one game. It was daytime the entire time :P. Now, it didn't actually say it was day which probably means an error happened somewhere, and I dont' know what that error was. But it wasn't kills failing. This was the message when you stopped the game: "Game lasted 31:47. 28:17 was day. 03:29 was night. "

PR #152, opened 2015-07-13
!stats Overhaul

skizzerz:
Decoupled ROLE_REVEAL from impacting !stats, added new STATS_TYPE.

ROLE_REVEAL is now a string:
- on = roles revealed on death
- off = roles not revealed on death
- team = user's team revealed on death but not their role

STATS_TYPE is also a string:
- default = stats are calculated as if a villager or nonplayer was
  manually tracking who was what; this leads to !stats not revealing
  any information that is not publicly known
- accurate = what it used to be with a couple modifications; all roles
  are shown (VG and time lord are no longer lumped with villager),
  alpha wolf and lycans turning are hidden, nothing else is hidden
  (so you can see what amnesiac turned into or when clone turns)
- team = only shows number of people on each team, calculated the same
  way as accurate
- disabled = stats are disabled, doing !stats only serves as a mass-ping
  of alive players

Removed random_reveal/random_noreveal hack
- !fgame random=role reveal:on|off does the same thing as the hack used
  to do.

Allow all game modes to take arguments (aka the blah in roles=blah).
- Allowed arguments are "role reveal" (on/off/team), "stats type"
  (default/accurate/team/disabled), and "abstain"
  (enabled/restricted/disabled).
- roles obviously still allows for taking individual roles as arguments
  as well as specifying a default role, however this functionality is
  not available for any other gamemode.
- if any custom options are set as a result of this or changing them in
  game mode constructors, players will be informed of which options are
  in play.

Performing !frole will force STATS_TYPE from default to accurate, this
is done because the default logic is based not on what roles are
currently in existence, but rather what roles the game was started with
and what transformations could have applied to them. !frole throws a
complete monkey wrench in that to the point where it is impossible to
nicely recover.

Clone and Traitor are now unconditionally blacklisted for amnesiac, as
the additional logic to support them in default !stats is incredibly
complex and as such was not done for this PR.

Matchmaker is conditionally blacklisted for amnesiac if amnesiacs turn
after the first night. It is an allowed choice (unless it appears on the
config blacklist) if amnesiacs turn night 1.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -102,6 +102,7 @@ var.OPPED = False  # Keeps track of whether the bot is opped
 
 var.BITTEN = {}
 var.BITTEN_ROLES = {}
+var.LYCAN_ROLES = {}
 var.VENGEFUL_GHOSTS = {}
 var.CHARMED = set()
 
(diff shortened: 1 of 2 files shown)


jacob1:
I really like this idea, it reduces the need to check !stats often, and adds some stuff that was hidden from !stats but obvious to the channel (like that the traitor exists). It will need some actual games to see how well it works and what other things might need to be added, but I think it is ready to be merged.

nyuszika7h:
@jacob1 How does that traitor thing even work?

src/wolfgame.py
```diff
@@ -1250,96 +1251,503 @@ def stats(cli, nick, chan, rest):
...
+            rolecounts["traitor"] = [0, 0]
+        # Step 2. Handle role swaps via exchange totem by modifying orig_roles -- the original
+        # roles themselves didn't change, just who has them. By doing the swap early on we greatly
+        # simplify the death logic below in step 3 -- to an outsider that doesn't know any info
+        # the role swap might as well never happened and those people simply started with those roles;
+        # they can't really tell the difference.
+        for a, b in var.EXCHANGED_ROLES:
+            temp = orig_roles[a]
+            orig_roles[a] = orig_roles[b]
+            orig_roles[b] = temp
```
nyuszika7h: This could be written as `(orig_roles[a], orig_roles[b]) = (orig_roles[b], orig_roles[a])` too.

src/wolfgame.py
```diff
@@ -1250,96 +1251,503 @@ def stats(cli, nick, chan, rest):
...
+                if count == 1:
+                    vb = "is"
+                else:
+                    vb = "are"
 
-        if role == rs[0]:
-            if count == 1:
-                vb = "is"
+            if count > 1 or count == 0:
+                if count == 0 and len(var.ORIGINAL_ROLES[role]) == 0:
```
nyuszika7h: This could be just `count != 1`, I guess it can't be negative.

src/wolfgame.py
```diff
@@ -1529,13 +1938,13 @@ def chk_decision(cli, force = ""):
...
                     if target != votee:
-                        if var.ROLE_REVEAL:
+                        if var.ROLE_REVEAL in ("on", "team"):
                             r1 = var.get_reveal_role(target)
                             an1 = "n" if r1.startswith(("a", "e", "i", "o", "u")) else ""
                             tmsg = ("As the noose is being fitted, \u0002{0}\u0002's totem emits a brilliant flash of light. " +
                                     "When the villagers are able to see again, they discover that \u0002{1}\u0002, " +
                                     "a{2} \u0002{3}\u0002, has fallen over dead.").format(votee, target, an1, r1)
-                        else:
+                        elif var.ROLE_REVEAL == "off":
```
nyuszika7h: It doesn't make a functional difference currently, but this is the only place you do this, in all the other places it's just `else:`.

src/wolfgame.py
```diff
@@ -4068,9 +4499,28 @@ def check_exchange(cli, actor, nick):
         var.ROLES[nick_role].append(actor)
         var.ROLES[nick_role].remove(nick)
         if actor in var.BITTEN_ROLES.keys():
-            var.BITTEN_ROLES[actor] = nick_role
-        if nick in var.BITTEN_ROLES.keys():
-            var.BITTEN_ROLES[nick] = actor_role
+            if nick in var.BITTEN_ROLES.keys():
+                temp = var.BITTEN_ROLES[actor]
+                var.BITTEN_ROLES[actor] = var.BITTEN_ROLES[nick]
+                var.BITTEN_ROLES[nick] = temp
```
nyuszika7h: Another place where you could use the `(a, b) = (b, a)` swap.

nyuszika7h:
Looks good, my comments are mostly minor style things.

PR #170, opened 2015-10-11
Require a minimum number of votes to start a game

lutoma:
Pretty much what it says in the title. Changes `wolfgame.py` to require at least `floor((num_players + 1) / 2.2)` to start, with an upper limit of 4.

Players | Votes<br />required
:---------:|:------------------:
4-5        | 2
6-10      | 3
>= 11    | 4

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -80,6 +80,9 @@ STATS_TYPE = "default" # default/accurate/team/disabled - what role information 
 LOVER_WINS_WITH_FOOL = False # if fool is lynched, does their lover win with them?
 DEFAULT_SEEN_AS_VILL = True # non-wolves are seen as villager regardless of the default role
 
+START_VOTES_SCALE = 0.3
+START_VOTES_MAX = 4
+
 # Debug mode settings, whether or not timers and stasis should apply during debug mode
 DISABLE_DEBUG_MODE_TIMERS = True
 DISABLE_DEBUG_MODE_TIME_LORD = False
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -439,6 +439,7 @@ def reset():
     var.NO_LYNCH = set()
     var.FGAMED = False
     var.GAMEMODE_VOTES = {} #list of players who have used !game
+    var.START_VOTES = set() # list of players who have voted to !start
     var.LOVERS = {} # need to be here for purposes of random
 
     reset_settings()
@@ -1164,7 +1165,7 @@ def join_player(cli, player, chan, who = None, forced = False):
             var.JOINED_THIS_GAME_ACCS.add(acc)
         var.CAN_START_TIME = datetime.now() + timedelta(seconds=var.MINIMUM_WAIT)
         cli.msg(chan, ('\u0002{0}\u0002 has started a game of Werewolf. '+
-                      'Type "{1}join" to join. Type "{1}start" to start the game. '+
+                      'Type "{1}join" to join. Type "{1}start" to vote to start the game. '+
                       'Type "{1}wait" to increase the start wait time.').format(player, botconfig.CMD_CHAR))
 
         # Set join timer
@@ -2185,6 +2186,11 @@ def show_votes(cli, nick, chan, rest):
         the_message = ", ".join(votelist)
         if len(pl) >= var.MIN_PLAYERS:
             the_message += "{0}Votes needed for a majority: {1}".format("; " if votelist else "", int(math.ceil(len(pl)/2)))
+
+        with var.WARNING_LOCK:
+            if var.START_VOTES:
+                the_message += "; Votes to start the game: {} ({})".format(len(var.START_VOTES), ', '.join(var.START_VOTES))
+
     elif var.PHASE == "night":
         cli.notice(nick, "Voting is only during the day.")
         return
@@ -2976,6 +2982,10 @@ def del_player(cli, nick, forced_death = False, devoice = True, end_game = True,
             if var.PHASE == "join":
                 if nick in var.GAMEMODE_VOTES:
                     del var.GAMEMODE_VOTES[nick]
+
+                with var.WARNING_LOCK:
+                    var.START_VOTES.discard(nick)
+
                 # Died during the joining process as a person
                 if var.AUTO_TOGGLE_MODES and nick in var.USERS and var.USERS[nick]["moded"]:
                     for newmode in var.USERS[nick]["moded"]:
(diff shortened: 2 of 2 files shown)


src/settings.py
```diff
@@ -80,6 +80,8 @@
 LOVER_WINS_WITH_FOOL = False # if fool is lynched, does their lover win with them?
 DEFAULT_SEEN_AS_VILL = True # non-wolves are seen as villager regardless of the default role
 
+START_VOTES_REQUIRED = lambda num_players: min((num_players + 1) / 2.2, 5)
```
Vgr255: If it's a function (even a lambda), make it lowercase, and preferably in the functions section. This will, among other things, allow us to segregate settings from functions in the future.

src/wolfgame.py
```diff
@@ -2161,6 +2162,9 @@ def show_votes(cli, nick, chan, rest):
         the_message = ", ".join(votelist)
         if len(pl) >= var.MIN_PLAYERS:
             the_message += "{0}Votes needed for a majority: {1}".format("; " if votelist else "", int(math.ceil(len(pl)/2)))
+
+        if len(var.START_VOTES) > 0:
```
Vgr255: Just make it `if var.START_VOTES`; there is no need to do a length check.

src/wolfgame.py
```diff
@@ -2161,6 +2162,9 @@ def show_votes(cli, nick, chan, rest):
...
+            the_message += "; Votes to start the game: {} ({})".format(len(var.START_VOTES), ', '.join(var.START_VOTES))
```
Vgr255: Make sure the semicolon is properly added only if there is a need to (see 3 lines up), else you could have a line beginning by that character.

src/wolfgame.py
```diff
@@ -6861,6 +6887,7 @@ def start(cli, nick, chan, forced = False, restart = ""):
         cgamemode(cli, restart)
         var.GAME_ID = time.time() # restart reaper timer
 
+    var.START_VOTES = set()
```
Vgr255: Move both this and line 84 into the reset() definition, so that it's properly reset if e.g. game times out or other stuff.

skizzerz:
This has been discussed before, and I still don't see the merit in it. What problems are you trying to solve by requiring multiple people to start? If you want to ensure everyone is around, having only a small handful of people say "start" isn't going to do anything about that. If you want to prevent "ninjastarters" this again will do nothing to prevent that, as there is no requirement that the starts need to be close enough to each other to actually be meaningful, and baking in that complexity is just adding in needless complexity for what is mostly a non-issue. The proper way to prevent "ninjastarting" is via social means, not technical. All I can really see this doing is ensuring games **don't** get started due to the reluctance to !start that I see in ##werewolf; invariably someone wants to wait for "that next person" even if it is quite obvious no such person is coming. So you start, then they wait a bunch of times because you need multiple people to start, so people go away, and you wind up with even more of a problem than you had to begin with.

Please elaborate on what issues you are hoping to solve with this PR, why you believe this PR solves them adequately, and why any downsides introduced by this system are less bad than the downsides without having the system.

skizzerz:
Per our discussion on IRC, I'd be willing to give this a shot for a while to see how it plays out, however I feel that it still needs some improvements before it can be merged.

Namely, I feel that there should be some sort of time limit to accrue all of the !starts, say 60 seconds from the first !start. If after 60 seconds there aren't enough, then the game does not start and the count resets to 0.

I also feel that after the initial !start, that we should either give different semantics to !wait (such as by deducting a start, essentially saying "no I'm not ready yet don't start now") or disallowing it entirely, as if there are 4 people that said "yes", it shouldn't be possible imo to delay it by another 30 seconds by a single person doing !wait.

src/wolfgame.py
lutoma: That can't really happen as it always at least prints "Votes needed for a majority: x" before (except for when there's not enough players, but then it also doesn't print the start votes).

src/wolfgame.py
```diff
@@ -6787,6 +6809,10 @@ def cgamemode(cli, arg):
         cli.msg(chan, "Mode \u0002{0}\u0002 not found.".format(modeargs[0]))
 
 
+def expire_start_votes(cli, chan):
```
skizzerz: Missing lock and phase check. Also that message is really awkward, I'd nix the bold at the very least.

src/wolfgame.py
```diff
@@ -6842,6 +6868,34 @@ def start(cli, nick, chan, forced = False, restart = ""):
...
+                remaining_votes = start_votes_required - len(var.START_VOTES)
+
+                if remaining_votes == 1:
+                    cli.msg(chan, msg.format(nick, remaining_votes, 'vote'))
+                else:
+                    cli.msg(chan, msg.format(nick, remaining_votes, 'votes'))
+
+                # If this was the first vote
+                if len(var.START_VOTES) == 1:
+                    with var.WARNING_LOCK:
```
skizzerz: lock in wrong place, this needs to be around anything that accesses var.START_VOTES (so add it in at line 6871 then indent all of this stuff by one)

src/settings.py
```diff
@@ -325,6 +326,7 @@
 PING_IF_NUMS_ACCS = {}
 
 is_role = lambda plyr, rol: rol in ROLES and plyr in ROLES[rol]
+start_votes_required = lambda num_players: min((num_players + 1) / 2.2, 4)
```
skizzerz: Prefer actual settings rather than lambdas-as-settings, even if it means you need to break it into more than one. Settings should really be scalars. Also, settings should be ALL_UPPERCASE. If this isn't meant to be a setting, don't put it in settings.py.

lutoma: It was originally all uppercase, but I turned it lowercase after some of the other devs told me to. I can change it again of course, though the other lambdas in settings.py are also lowercase.

Regarding turning it into a variable, I'm not so sure. I guess I could turn it into a simple variable of a percentage of users, but that wouldn't allow more "complex" calculations like the current one and requires a more akward syntax to completely disable it (I guess you could set it to 0%).

skizzerz: The other lambdas in settings.py shouldn't be in settings.py because they aren't settings. Legacy code doesn't give an excuse for new code to do dumb things, however. This lambda should instead be hardcoded into the function that actually runs it (or if more than one function runs it, then move it into wolfgame.py as-is so we don't needlessly duplicate code).

In terms of making it configurable, you'd use two variables -- var.START_VOTES_SCALE and var.START_VOTES_MAX. I'd also reconfigure the algorithm to act more like gunner bullet determination, so `min(ceil(num_players * var.START_VOTES_SCALE), var.START_VOTES_MAX)`. Doing it that way allows flexibility in the scaling as well as easily disabling it entirely if someone doesn't want it. Default values would be var.START_VOTES_SCALE = 0.4 and var.START_VOTES_MAX = 4, which retains the table you specified above.

PR #177, opened 2015-10-18
Group all night messages together

### LINKED ISSUE #174: Player list is sent twice for wolf-assassins
When both a wolf-character and an assassin, lykos seems to send the list of players twice, once for the wolf role (with wolfteam-aligned roles revealed), and once for assassin, without anything revealed.

That wall of text makes it really easy to miss the fact you're also assassin. Instead, it should probably only print the players with roles once after the other messages.

```
You are a hag. You can hex someone to prevent them from using any special powers they may have during the next day and night. Use "hex <nick>" to hex them. Only detectives can reveal your true identity, seers will see you as a regular villager.
Also, if you PM me, your message will be relayed to other wolves.
Players: banzaikitten, abra0, sapphire (wolf), m712, Pichu, fetergent (cursed traitor), maserati (wolf), Feeshee, AvatarFreak24, mellowcraze, michagogo (cursed), jacob1, nyuszika7h
You are an assassin. Choose a target with "target <nick>". If you die you will take out your target with you. If your target dies you may choose another one.
Players: AvatarFreak24, banzaikitten, Pichu, jacob1, m712, michagogo, maserati, Feeshee, abra0, nyuszika7h, sapphire, fetergent, mellowcraze
```

Vgr255:
This groups messages together for purposes of role + assassin and some potential future template. I have purposefully left minion out, as the role does not quite fit the same rules as the others. It will send two different lists if they are also assassin, but we want that, NOT the opposite! Doctor also now gets a player list, which may very well be outdated by daybreak, but detective gets player list too so why not doctor.

Edit: This fixes #174

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6227,7 +6227,8 @@ def transition_night(cli):
         return
 
     # send PMs
-    ps = var.list_players()
+    ps = set(var.list_players())
+    to_send = defaultdict(lambda: ["Players", set()])
     wolves = var.list_players(var.WOLFCHAT_ROLES)
     for wolf in wolves:
         normal_notify = wolf in var.PLAYERS and not is_user_simple(wolf)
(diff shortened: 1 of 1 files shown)


Vgr255:
To preserve mergeability of some of my modifications, this will need to be closed and re-opened only when Halloween has come and gone.

PR #191, opened 2015-11-11
Introduce proxies as a means of breaking circular import chains

skizzerz:
What it says on the tin

### DIFF
--- a/src/proxy.py
+++ b/src/proxy.py
@@ -3,21 +3,26 @@ import inspect
 from src.decorators import handle_error
 
 """ This module introduces two decorators - @proxy.stub and @proxy.impl
+
 @proxy.stub is used to decorate a stub method that should be filled in
 with an implementation in some other module. Calling the stub method
 calls the implementation method in the other module instead of the stub
 method itself (or raises a NotImplementedError if no such
 implementation exists)
+
 @proxy.impl is used to define a previously-declared stub with an actual
 implementation to call -- the signature for the implementation must
 exactly match the signature for the stub (enforced for Python 3.3+).
+
 Attempting to implement a non-existent stub is an error, as is trying
 to re-implement a stub that is already implemented.
+
 Example:
 (foo.py)
 @proxy.stub
 def my_method(foo, bar=10):
     pass
+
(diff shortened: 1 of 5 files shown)


skizzerz:
Don't merge yet: should support multiple proxy stubs for a single implementation, provided all the stubs have the same signature (e.g. gamemodes.py can declare a stub for something, and roles.py can declare a stub for something, and wolfgame.py can define the implementation of that and it should work, provided the two stubs share the same signature).

src/proxy.py
```diff
@@ -0,0 +1,96 @@
+import inspect
+
+from src.decorators import handle_error
```
Vgr255: Seems to be an artefact; it's not used.

src/proxy.py
```diff
@@ -0,0 +1,96 @@
...
+        SIGS[f.__name__] = inspect.signature(f)
+    else:
+        # Python 3.2; not allowed to have nice things
+        SIGS[f.__name__] = None
+
+    return inner
+
+def impl(f):
+    if f.__name__ not in SIGS:
+        raise NameError(("Attempting to implement a proxy stub {0} that does "
```
Vgr255: Nit: Use `{0!r}` to have quotes around the name; makes it more readable imo.

src/wolfgame.py
```diff
@@ -2074,6 +2067,7 @@ def fday(cli, nick, chan, rest):
 
 # Specify force = "nick" to force nick to be lynched
 @handle_error
```
Vgr255: Can remove that, `impl` already wraps.

src/proxy.py
Vgr255: Also bunch of other places where this'd nicer-looking.

Vgr255:
Couple of nitpicks pointed out, other than that looks fine.

src/wolfgame.py
skizzerz: No it doesn't. it wraps it for things called using the stub, but the original method is not wrapped in impl.

Vgr255: Oh, right.

PR #200, opened 2015-11-30
Only allow retracting start votes in the main channel

lutoma:


skizzerz:
Agreed with nyuszika7h; the modus operandi for lykos has been largely transitioned to silent failures if commands are no-ops in a given context. The fact that lykos doesn't publicly announce that the vote was retracted should be enough to clue the user into the fact they need to issue the command in-channel. Looking even at the same command, the user is not told they need to !retract in channel for a !vote (nor should they be as it is obvious whether or not the command works).

PR #201, opened 2015-12-04
Fixes #184 (!swap, when targeting someone, should be case insensitive)

### LINKED ISSUE #184: !swap is sensitive to capitalization
[20:59] <NotMJAtAll> swap mj94
[20:59] <lykos> That person is not playing.
[20:59] <NotMJAtAll> swap MJ94
[20:59] <lykos> You are a villager.

Lykos should support any nick capitalization, as other commands such as !lynch and !vote do.

Diitto:
:)

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -767,7 +767,15 @@ def replace(cli, nick, chan, rest):
                 cli.notice(nick, msg)
             return
     else:
-        target, _ = complete_match(rest[0], var.list_players() + list(var.VENGEFUL_GHOSTS.keys()))
+        pl = var.list_players() + list(var.VENGEFUL_GHOSTS.keys())
+        pll = [var.irc_lower(i) for i in pl]
+
+        lc_target, _ = complete_match(var.irc_lower(rest[0]), pll)
+
+        if lc_target != None:
+            target = pl[pll.index(lc_target)]
+        else:
+            target = ""
 
         if target not in var.list_players() and target not in var.VENGEFUL_GHOSTS:
             msg = messages["target_not_playing"].format(" longer" if target in var.DEAD else "t")


Diitto:
:(

src/wolfgame.py
```diff
@@ -767,7 +767,11 @@ def replace(cli, nick, chan, rest):
                 cli.notice(nick, msg)
             return
     else:
-        target, _ = complete_match(rest[0], var.list_players() + list(var.VENGEFUL_GHOSTS.keys()))
+        plyr_list = var.list_players + list(var.VENGEFUL_GHOSTS.keys())
+        lc_list   = [i.lower() for x in plyr_list]
```
skizzerz: Use `pl` for player list and `pll` for lowercased player list for consistency with rest of lykos code. Also, please use IRC lowercasing instead of str.lower() -- I know existing code incorrectly uses str.lower() but that doesn't give an excuse for new code to. The IRC lowercasing function is in src/settings.py iirc (aka `var`).

Also, **test your code** before submitting a PR next time; it's pretty obvious that you didn't test this at all given the egregious error in it: **i**.lower() for **x**... A PR should only be submitted if whatever is changed has actually been verified to work and not blow up.

src/wolfgame.py
```diff
@@ -767,7 +767,11 @@ def replace(cli, nick, chan, rest):
                 cli.notice(nick, msg)
             return
     else:
-        target, _ = complete_match(rest[0], var.list_players() + list(var.VENGEFUL_GHOSTS.keys()))
+        pl = var.list_players() + list(var.VENGEFUL_GHOSTS.keys())
+        pll = [var.irc_lower(i) for i in pl]
+        
+        lc_target, _ = complete_match(var.irc_lower(rest[0]), pll)
+        target = pl[pll.index(lc_target)]
```
skizzerz: breaks if complete_match was unable to find a match (lc_target is None in that case, and index() raises an exception)

src/wolfgame.py
```diff
@@ -767,7 +767,15 @@ def replace(cli, nick, chan, rest):
                 cli.notice(nick, msg)
             return
     else:
-        target, _ = complete_match(rest[0], var.list_players() + list(var.VENGEFUL_GHOSTS.keys()))
+        pl = var.list_players() + list(var.VENGEFUL_GHOSTS.keys())
+        pll = [var.irc_lower(i) for i in pl]
+
+        lc_target, _ = complete_match(var.irc_lower(rest[0]), pll)
+
+        if lc_target != None:
```
skizzerz: use `is not None` here and do not replace target for the else condition -- otherwise the error message is very unhelpful (aka remove the else condition)

jacob1:
merged

PR #203, opened 2015-12-04
Idle out players in join phase in lieu of canceling the whole game after a certain time

lutoma:


src/settings.py
```diff
@@ -44,6 +43,8 @@
 KILL_IDLE_TIME = 300
 WARN_IDLE_TIME = 180
 PM_WARN_IDLE_TIME = 240
+JOIN_PHASE_KILL_IDLE_TIME = 1800
+JOIN_PHASE_WARN_IDLE_TIME = 1500
```
skizzerz: way too long imo, I'd warn after 15 minutes and kill after 20 at the most; this really needs to be discussed though -- feel free to prod -ops to kick that off.

src/wolfgame.py
```diff
@@ -3167,15 +3155,25 @@ def reaper(cli, gameid):
     var.IDLE_WARNED_PM = set()
     chan = botconfig.CHANNEL
 
-    last_day_id = var.DAY_COUNT
+    last_day_id = getattr(var, 'DAY_COUNT', None)
```
skizzerz: Use double quotes for strings, also prefer to just set var.DAY_COUNT to 0 in reset() rather than using getattr(). The double quotes for strings applies for every other instance you used single quotes as well.

src/wolfgame.py
```diff
@@ -3189,56 +3187,86 @@ def reaper(cli, gameid):
                         var.LAST_SAID_TIME[nick] += timedelta(seconds=10*num_night_iters)
                     num_night_iters = 0
 
-
-            if not skip and (var.WARN_IDLE_TIME or var.PM_WARN_IDLE_TIME or var.KILL_IDLE_TIME):  # only if enabled
+            if not skip and idle_timeouts_enabled:  # only if enabled
```
skizzerz: kill the comment, it doesn't explain anything you couldn't already determine from the code itself (I know you didn't add the comment, but you're changing the line anyway :P)

src/wolfgame.py
```diff
@@ -7643,12 +7674,6 @@ def start(cli, nick, chan, forced = False, restart = ""):
             if var.STASISED_ACCS[acc] <= 0:
                 del var.STASISED_ACCS[acc]
 
-    if not botconfig.DEBUG_MODE or not var.DISABLE_DEBUG_MODE_REAPER:
-        # DEATH TO IDLERS!
-        reapertimer = threading.Thread(None, reaper, args=(cli,var.GAME_ID))
-        reapertimer.daemon = True
-        reapertimer.start()
-
```
skizzerz: The reaper thread only lasts for a single game (e.g. while var.GAME_ID remains unchanged). var.GAME_ID changes both when a game is first created via !join as well as when a game is !started. So removing this means you just disabled the in-game reaper, nice job breaking it hero.

MJ94:
Will this check a user's /whois idle time or channel chat idle time? Preferably the former.

skizzerz:
Channel chat, this is to keep it consistent with the in-game reaper and is actually feasible. Checking whois idle time varies by ircd and may be entirely impossible for some, not to mention the gigantic amount of lag amassed by issuing a ton of whois queries.

MJ94:
Right, but I often idle in the channel when I'm waiting for a game to start, so we'd want to make sure it doesn't just assume I'm idle and quit me without checking first, for example.

skizzerz:
It would give you a message similar to the one in-game that pings you 5 minutes prior to it quitting you for idling. Assuming you're somewhat paying attention to pings, you should be able to reply in-channel during those 5 minutes.

skizzerz:
Closing due to inactivity. If you are still interested in this, please reopen the PR after fixing the mentioned issues.

PR #207, opened 2015-12-07
Fix first night message

lutoma:
This got mixed up in the whole move of the messages to an external file. Right now it always prints the first night message («… If you did not receive one, simply sit back, relax, and wait patiently for morning.») on every night _except_ for the first night. This reverts it to the proper behaviour.


Vgr255:
This is intentional behaviour.

PR #208, opened 2015-12-07
Support IPv6 connections (now with more TLS)

lutoma:


oyoyo/client.py
```diff
@@ -177,13 +174,17 @@ def connect(self):
...
                 try:
-                    self.socket.connect(("{0}".format(self.host), self.port))
+                    self.socket = socket.create_connection(("{0}".format(self.host), self.port))
                     break
                 except socket.error as e:
                     retries += 1
                     self.stream_handler('Error: {0}'.format(e), level="warning")
                     if retries > 3:
-                        break
+                        exit(os.EX_UNAVAILABLE)
```
Vgr255: The os module doesn't have such an attribute; it's probably platform-dependent (or just doesn't exist). Please use whatever constant it corresponds to, or just `1`.

nyuszika7h: Yeah, it [https://docs.python.org/3.4/library/os.html#os.EX_UNAVAILABLE](seems) it's Unix-specific, but an exit code of 69 is weird anyway. And you should use **sys**.exit in scripts.

PR #210, opened 2015-12-22
Added a few death messages

Cr0wb4r:
They don’t have to be used, but I thought it might be fun to have a few
more, and a bit more variety.


jacob1:
This is based off of an old version of the code and is unmergeable. Maybe try getting the latest source and add the messages back, near the top of this file: https://github.com/lykoss/lykos/blob/master/messages/en.json

Also, left overs -> leftovers :)

nyuszika7h:
Also, Roulete → Roulette.

PR #218, opened 2016-04-04
Add CTCP VERSION support

io4:
Now the bot will try to get latest commit short id, and if that fails, don't include it.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6567,7 +6567,7 @@ def relay(cli, nick, chan, rest):
         try:
             ans = subprocess.check_output(["git", "log", "-n", "1", "--pretty=format:%h"])
             reply = "\u0001VERSION lykos {0}, Python {1} -- https://github.com/lykoss/lykos\u0001".format(str(ans.decode()), platform.python_version())
-        except CalledProcessError:
+        except (OSError, subprocess.CalledProcessError):
             reply = "\u0001VERSION lykos, Python {0} -- https://github.com/lykoss/lykos\u0001".format(platform.python_version())
         cli.notice(nick, reply)
         return
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6566,8 +6566,8 @@ def relay(cli, nick, chan, rest):
     if rest == "\u0001VERSION\u0001":
         try:
             ans = subprocess.check_output(["git", "log", "-n", "1", "--pretty=format:%h"])
-            reply = "\u0001VERSION lykos {0}, Python {1} -- https://github.com/lykoss/lykos\u0001".format(str(ans.decode()),platform.python_version())
-        except:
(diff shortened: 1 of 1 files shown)


src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
     if rest.startswith("\u0001PING"):
         cli.notice(nick, rest)
         return
+    if rest.startswith("\u0001VERSION"):
+        try:
+          ans=subprocess.check_output(["git", "log", "-n", "1", "--pretty=format:%h"])
```
skizzerz: Formatting: needs spaces around the = sign

src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
...
+          reply = "lykos "+str(ans.decode())+", python v"+platform.python_version()+" -- https://github.com/lykoss/lykos"
```
skizzerz: use str.format instead of string concatenation, capitalize Python, drop the "v" in front of the python version (same with below in the except)

src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
     if rest.startswith("\u0001PING"):
         cli.notice(nick, rest)
         return
+    if rest.startswith("\u0001VERSION"):
```
Vgr255: Make that `rest == "\u0001VERSION\u0001"` instead.

src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
...
+        try:
+          ans=subprocess.check_output(["git", "log", "-n", "1", "--pretty=format:%h"])
+          reply = "lykos "+str(ans.decode())+", python v"+platform.python_version()+" -- https://github.com/lykoss/lykos"
+        except:
+            reply = "lykos, python v"+platform.python_version()+" -- https://github.com/lykoss/lykos"
+        cli.notice(nick, reply)
```
skizzerz: reply needs to begin with the string VERSION

src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
     if rest.startswith("\u0001PING"):
         cli.notice(nick, rest)
         return
+    if rest == "\u0001VERSION\u0001":
+        try:
+            ans = subprocess.check_output(["git", "log", "-n", "1", "--pretty=format:%h"])
+            reply = "\u0001VERSION lykos {0}, Python {1} -- https://github.com/lykoss/lykos\u0001".format(str(ans.decode()),platform.python_version())
```
skizzerz: Formatting: space after comma

src/wolfgame.py
```diff
@@ -6562,6 +6563,14 @@ def relay(cli, nick, chan, rest):
...
+        except:
```
skizzerz: except too broad, should only catch exceptions from subprocess.check_output, e.g. CalledProcessError

PR #219, opened 2016-04-07
Notify players in deadchat when the game ends

cyberfawkes:
Added a PM to inform Deadchat Players that the game has ended.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -2440,8 +2440,6 @@ def stop_game(cli, winner = "", abort = False, additional_winners = None):
 
     if not abort:
         cli.msg(chan, gameend_msg)
-        # Message Players in Deadchat letting them know that the game has ended. cyberfawkes
-        mass_privmsg(cli, var.DEADCHAT_PLAYERS, messages["endgame_deadchat"].format(chan))
         
     roles_msg = []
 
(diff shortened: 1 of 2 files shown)


src/wolfgame.py
```diff
@@ -2440,7 +2440,9 @@ def stop_game(cli, winner = "", abort = False, additional_winners = None):
 
     if not abort:
         cli.msg(chan, gameend_msg)
-
+        # Message Players in Deadchat letting them know that the game has ended. cyberfawkes
+        mass_privmsg(cli, var.DEADCHAT_PLAYERS, messages["endgame_deadchat"].format(chan))
```
skizzerz: This should be moved to after the end-of-game stuff has been run, but before reset(). Displaying the list of winners and recording the stats is more important than letting the people in deadchat know the game is over. Also, remove your name from the comment. Nobody "owns" the code, and git already supplies attribution information.

nyuszika7h: Nitpick: please Get rid Of unnecessary Capitalization. :P

cyberfawkes: I'll fix it right now, I forgot to remove my name. I left it in there for debugging so I didn't have to remember the specific line that line of code was on, I could just search "cyberfawkes".

nyuszika7h:
Merged as 71ff8c3.

PR #221, opened 2016-04-10
Add the thief role

Vgr255:
This adds the new thief role into lykos. Entirely untested for now because I'm lazy. Will test soon-ish, but open to reviews before then.

Don't merge yet, I want to introduce a game mode along with it, but I haven't got around to it yet.


skizzerz:
As mentioned in IRC, the design of this is flawed and requires massive kludges in order to work somewhat properly. Either the roles should be fully swapped before kills (including choices the old role made, e.g. if thief steals GA they also steal who that GA guarded), or fully swapped after kills (choices don't matter in this case). Kinda-sorta swapping before but actually performing the swap after requires massive hacks, and what is in the PR barely scratches the surface of what is required. The already-bad transition_day and del_player functions would be made pretty much unreadable by the level of hackery required to fully support thief in a half-half solution.

Vgr255:
Will do this again when the roles are completely moved over.

PR #222, opened 2016-04-17
Lone wolf

rdococ:
I'm not very good with this whole pull request thing... Anyway, the lone wolf is basically a wolf that fights for itself, and doesn't vote with the other wolves on who to kill. It has to survive to the end of the game to win.

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -258,13 +258,17 @@ ROLE_GUIDE = OrderedDict([ # This is order-sensitive - many parts of the code re
              ("mayor"            , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  1  ,  1  )),
              ("assassin"         , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  )),
              ("bureaucrat"       , (  0  ,  0  ,  0  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  )),
+             ("lone wolf"        , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  )),
              ])
+
+# Counts as a wolf in the win check, but doesn't necessarily win with wolves - opposite of WOLFTEAM_ROLES
+CHKWOLF_ROLES = {"lone wolf"}
 
 # Harlot dies when visiting, seer sees as wolf, gunner kills when shooting, GA and bodyguard have a chance at dying when guarding
 # If every wolf role dies, and there are no remaining traitors, the game ends and villagers win (monster may steal win)
-WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})
+WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})# | LONEWOLF_ROLES
 # Access to wolfchat, and counted towards the # of wolves vs villagers when determining if a side has won
-WOLFCHAT_ROLES = WOLF_ROLES | {"traitor", "hag", "sorcerer", "warlock"}
+WOLFCHAT_ROLES = (WOLF_ROLES) | {"traitor", "hag", "sorcerer", "warlock"}
 # Wins with the wolves, even if the roles are not necessarily wolves themselves
 WOLFTEAM_ROLES = WOLFCHAT_ROLES | {"minion", "cultist"}
 # These roles either steal away wins or can otherwise win with any team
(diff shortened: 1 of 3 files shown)


src/wolfgame.py
```diff
@@ -6918,7 +6920,7 @@ def transition_night(cli):
         if role == "wolf mystic":
             # if adding this info to !myrole, you will need to save off this count so that they can't get updated info until the next night
             # # of special villagers = # of players - # of villagers - # of wolves - # of neutrals
-            numvills = len(ps) - len(var.list_players(var.WOLFTEAM_ROLES)) - len(var.list_players(("villager", "vengeful ghost", "time lord", "amnesiac", "lycan"))) - len(var.list_players(var.TRUE_NEUTRAL_ROLES))
+            numvills = len(ps) - len(var.list_players(var.WOLFTEAM_ROLES | var.CHKWOLF_ROLES)) - len(var.list_players(("villager", "vengeful ghost", "time lord", "amnesiac", "lycan"))) - len(var.list_players(var.TRUE_NEUTRAL_ROLES))
```
Vgr255: `var.CHKWOLF_ROLES` is a very bad (and undescriptive) name. Please consider something a bit less cryptic.

src/wolfgame.py
```diff
@@ -4766,7 +4766,7 @@ def chk_nightdone(cli):
                            "sorcerer", "hunter", "hag", "shaman", "crazed shaman",
                            "augur", "werekitten", "warlock", "piper", "wolf mystic",
                            "fallen angel", "vigilante", "doomsayer", "doomsayer", # NOT a mistake, doomsayer MUST be listed twice
-                           "prophet", "wolf shaman", "wolf shaman") # wolf shaman also must be listed twice
+                           "prophet", "wolf shaman", "wolf shaman", "lone wolf") # wolf shaman also must be listed twice Modified
```
Vgr255: This "Modified" comment has no purpose there. Also please put lone wolf before both occurrences of wolf shaman, so that the comment makes sense.

src/settings.py
```diff
@@ -258,21 +258,25 @@
              ("mayor"            , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  1  ,  1  )),
              ("assassin"         , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  )),
              ("bureaucrat"       , (  0  ,  0  ,  0  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  ,  1  )),
+             ("lone wolf"        , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  )),
              ])
 
+# Counts as a wolf in the win check, but doesn't necessarily win with wolves - opposite of WOLFTEAM_ROLES
+CHKWOLF_ROLES = {"lone wolf"}
```
Vgr255: Besides the bad name, this goes with the other similar sets, and this should be a frozenset.

src/settings.py
```diff
@@ -258,21 +258,25 @@
...
+
 # Harlot dies when visiting, seer sees as wolf, gunner kills when shooting, GA and bodyguard have a chance at dying when guarding
 # If every wolf role dies, and there are no remaining traitors, the game ends and villagers win (monster may steal win)
-WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})
+WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})# | LONEWOLF_ROLES
 # Access to wolfchat, and counted towards the # of wolves vs villagers when determining if a side has won
-WOLFCHAT_ROLES = WOLF_ROLES | {"traitor", "hag", "sorcerer", "warlock"}
+WOLFCHAT_ROLES = (WOLF_ROLES) | {"traitor", "hag", "sorcerer", "warlock"}
```
nyuszika7h: Why did you add the parentheses here?

src/settings.py
```diff
@@ -258,21 +258,25 @@
...
+             ("lone wolf"        , (  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  ,  0  )),
              ])
 
+# Counts as a wolf in the win check, but doesn't necessarily win with wolves - opposite of WOLFTEAM_ROLES
+CHKWOLF_ROLES = {"lone wolf"}
+
 # Harlot dies when visiting, seer sees as wolf, gunner kills when shooting, GA and bodyguard have a chance at dying when guarding
 # If every wolf role dies, and there are no remaining traitors, the game ends and villagers win (monster may steal win)
-WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})
+WOLF_ROLES = frozenset({"wolf", "alpha wolf", "werecrow", "wolf cub", "werekitten", "wolf mystic", "wolf shaman", "fallen angel", "doomsayer"})# | LONEWOLF_ROLES
```
nyuszika7h: Remove the comment, we don't need dead code in the repo.

skizzerz:
What does this role do? How does it win? How does it interact with other roles? etc.

Your PR should explain all of this, so please add that information as a reply.

skizzerz:
This is a horrible idea (same goes for what I understand white wolf to be, so I'll address the issue of a wolfteam-member-except-not in general here).

I'm not entirely opposed to roles that are wolves or in wolfchat but have ulterior motives. However, those motives **cannot** clash with how wolves in general operate. The bot makes a lot of assumptions whenever you put something into WOLF_ROLES and WOLFCHAT_ROLES (notably, the former must also be a subset of the latter). These assumptions determine how these role types interact with other roles in general to ensure a consistent experience among them without needing to modify 20 places every time a wolf is added. In addition to technical assumptions, there is a social assumption that everyone in wolfchat is superficially working together. Anyone in wolfchat that has ulterior motives must still be working towards the greater goal of wolfteam in order for wolfchat to work, otherwise the wolf team completely falls apart because an unknown-but-untrusted user has full access to everything they do (see also: hag lover). In this case, having the lone wolf select their kill individually is an instant red flag for the rest of the wolves and removes any secrecy, and as a wolf role they must be in wolfchat. Making a wolf not in wolfchat would be incredibly hacky and I would be opposed to it due to how much it would uglify the code -- while not relaying things they do is easy enough, you need to consider how they interact with the rest of wolfteam if they aren't in wolfchat; e.g. can other wolves kill them, can they kill other wolves, if not you have an info leak and if so you've made the game nearly unwinnable for wolves unless they choose to collaborate and therefore violate the entire premise of the role.

An example of the social aspect working properly would be a wolf role that only wins if wolves win and they are the last wolf standing. They still work with the other wolves in selecting kills, and offer no reason for the other wolves to suspect that they are working behind their backs. Even so, such a role wouldn't really fit into lykos (or rather, how freenode plays lykos) as in order for such a role to win, they need to be feeding the wolf list to the safes, while somehow expecting themselves to stay alive.

You can attempt to convince me otherwise, but you will need to bring up a good argument to do so (read: not a single sentence, dig deep into how the role as posed can successfully interact with wolves and other roles, what niche they fill, and how this doesn't completely break the bot or how the game is played both in freenode's setting as well as a setting that more matches IRL games). If I don't hear anything back by next week, I'm closing this PR.

skizzerz:
Closing per above comment.

PR #225, opened 2016-06-07
Revamp database schema and permissions backend

skizzerz:
This PR hits on a number of things that are all mostly related:

- New database schema. This impacts how player/game stats are stored as well as other things such as preferences and stasis.
- As a consequence, all stats are effectively reset; the old stats still exist but are not queried by the !pstats or !gstats commands (nor should they be)
- The new schema is capable of associating multiple player records (accounts and hostmasks) with a single person. Things like stats are attached to players whereas warnings, stasis, and preferences are attached to people. Viewing and manipulating these mappings via the bot is currently not possible and will not be part of this PR. A future change may add in this functionality, otherwise these can be viewed and manipulated by directly querying the database.
- Break admin commands into numerous groups, each associated with a flag (think IRC access levels). Full admins are +F and can do everything, otherwise groups can be assigned to people using !fflag. This replaces !fallow, which no longer exists (allowing commands on a per-command basis is no longer possible; as part of migration anyone previously allowed commands is now granted the relevant flags automatically). This also replaces the ADMINS and ADMINS_ACCOUNTS variables in botconfig, which are no longer used.
- Owner checking is unchanged and owners must still be defined in botconfig
- !fstasis can now only decrease stasis; it cannot increase it. The !fwarn command should be used to add stasis to someone instead.
- Stasis now expires automatically. Each time stasis is given, the expiration is increased by 1 hour. After the expiration timer is passed, all stasis expires at once. For example, giving 3 stasis means that all 3 (or whatever is left of them) expire after 3 hours, it does NOT expire at a rate of 1/hour.
- The !fwarn command was added in order to give warnings to users. These warnings can have a public reason (shown to the user), private notes (shown only to admins), and various sanctions (such as requiring acknowledgement before the user can !join again, giving out stasis, or denying commands). This replaces !fdeny, which no longer exists; as part of migration any previously denied commands are now converted into 0-point warnings which never expire and deny those specific commands.
- When awarding warning points, various sanctions may automatically apply even if the warning does not specify them. These operate on a threshold basis, in other words crossing that warning point threshold applies that sanction. By default, 1-4 warning points require acknowledgement, 5-9 each give one game of stasis, 10 requires acknowledgement and gives 3 games of stasis, 11-14 each give three games of stasis, and 15-24 each give an increasing amount of stasis (5-14 games). 25 is set to temporarily ban the user until enough warnings expire to drop them to 15 points or below, however tempban functionality is not yet implemented and will not be implemented as part of this PR.
- Due to the automatic sanctions, it is generally not necessary to specify any yourself; simply set the appropriate number of warning points. Sanctions can still be manually specified if desired, however.
- The !warn command can be used to view your own warnings and acknowledge them.
- !fwarn can modify the reason and notes for an existing warning, but cannot modify the warning points, expiration, or sanctions; fwarn can delete warnings however which makes them stop taking effect.


skizzerz:
http://termbin.com/wdba http://termbin.com/mpwj

jacob1:
The main issue I've noticed is that you can't join games ... it tells me I have unacknowledged warnings when I don't. has_unacknowledged_warnings is returning True for some reason.

This next thing I think is an issue (but it could just be it refusing to let me join the game still). If you have an expired but unacknowledged warning, it won't let you join the game and will tell you to !warn list. But it won't appear in the list unless you do -all. It doesn't tell you to do -all anywhere. Probably it should show unacknowledged anyway even if they are expired.

Minor bug:
Mon 06-13 [20:44:04] <jacob1> !!fwarn set
Mon 06-13 [20:44:05] <wofljacobot> Usage: fwarn del <id>

Is there a way to change the expiration time on a warning? If I accidentally make one permanent or want to make permanent one expire but still show up in a user's !warn list -all, then there should probably be a way to do that.

messages/en.json
```diff
@@ -796,6 +797,69 @@
...
+    "warn_ack_syntax": "Usage: warn ack <id>",
+    "warn_help_syntax": "Uwage: warn help <subcommand>",
+    "fwarn_add_syntax": "Usage: fwarn add <nick[!user@host]|=account> [@]<points> [~expiry] [sanctions] <:reason> [| notes]",
+    "fwarn_page_invalid": "Invalid page, must be a number 1 or greater.",
+    "fwarn_points_invalid": "Invalid points, must be a number above 0.",
+    "fwarn_expiry_invalid": "Invalid expiration amount, must be a number above 0 or 'never' for a warning that never expires.",
+    "fwarn_expiry_invalid_suffix": "Invalid expiration suffix, must use either d, h, or m.",
+    "fwarn_cannot_add": "Cannot add warning, double-check your parameters (the nick might be wrong or you are not joined to the channel).",
+    "fwarn_added": "Added warning {0}.",
+    "fwarn_done": "Done.",
```
Vgr255: Seems a bit odd to put this as "fwarn_done" - what about just "done" and put that somewhere a bit more generic?

messages/en.json
```diff
@@ -796,6 +797,69 @@
...
+    "fwarn_sanction_invalid": "Invalid sanction, can be either deny or stasis.",
+    "fwarn_stasis_invalid": "Invalid stasis amount, specify sanction as stasis=number.",
+    "fwarn_deny_invalid": "Invalid denied commands, specify sanction as deny=command,command,command without spaces.",
```
Vgr255: "command without spaces" ... ?

nyuszika7h: It's meant to be read as `command, command, command` without spaces.

On 16 June 2016 22:59:23 CEST, Emanuel Barry <notifications@github.com> wrote:
>> +    "fwarn_help_syntax": "Usage: fwarn help <subcommand>",
>> +    "warn_list_syntax": "Usage: warn list [-all] [page]",
>> +    "warn_view_syntax": "Usage: warn view <id>",
>> +    "warn_ack_syntax": "Usage: warn ack <id>",
>> +    "warn_help_syntax": "Uwage: warn help <subcommand>",
>> +    "fwarn_add_syntax": "Usage: fwarn add
><nick[!user@host]|=account> [@]<points> [~expiry] [sanctions] <:reason>
>[| notes]",
>> +    "fwarn_page_invalid": "Invalid page, must be a number 1 or
>greater.",
>> +    "fwarn_points_invalid": "Invalid points, must be a number above
>0.",
>> +    "fwarn_expiry_invalid": "Invalid expiration amount, must be a
>number above 0 or 'never' for a warning that never expires.",
>> +    "fwarn_expiry_invalid_suffix": "Invalid expiration suffix, must
>use either d, h, or m.",
>> +    "fwarn_cannot_add": "Cannot add warning, double-check your
>parameters (the nick might be wrong or you are not joined to the
>channel).",
>> +    "fwarn_added": "Added warning {0}.",
>> +    "fwarn_done": "Done.",
>> +    "fwarn_sanction_invalid": "Invalid sanction, can be either deny
>or stasis.",
>> +    "fwarn_stasis_invalid": "Invalid stasis amount, specify sanction
>as stasis=number.",
>> +    "fwarn_deny_invalid": "Invalid denied commands, specify sanction
>as deny=command,command,command without spaces.",
>
>"command without spaces" ... ?
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/lykoss/lykos/pull/225/files/16a0a28e360087b90b0640c724fe2b0039f81697#r67423006

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Vgr255: Oh, yeah. Should probably add quotes and/or backticks (but if you're going to do this, do that for all the messages).

messages/en.json
```diff
@@ -796,6 +797,69 @@
...
+    "fwarn_deny_invalid_command": "Invalid command \"{0}\", specify sanction as deny=command,command,command without spaces.",
+    "fwarn_list_header": "{0} has {1} active warning points. Warnings prefixed with \u0002!\u0002 are unacknowledged.",
+    "warn_list_header": "You have {0} active warning points. You must acknowledge all warnings prefixed with \u0002!\u0002 by using \"warn ack <id>\" before you can join games.",
+    "fwarn_list": "{0}{1}[#{2} {3}] to {4} by {5} - {6} ({7} points, {8}){9}",
+    "warn_list": "{0}{1}[#{2} {3}] {4} ({5} points, {6}){7}",
+    "fwarn_deleted": "deleted",
+    "fwarn_expired": "expired",
+    "fwarn_list_expired": "expired on {0}",
```
Vgr255: Eh, I'd say probably same as above, but this happens to be very specific to fwarn, so whatever.

messages/en.json
```diff
@@ -796,6 +797,69 @@
...
+    "fwarn_never_expires": "never expires",
+    "fwarn_list_footer": "More results are available, use fwarn list {0} to view them.",
+    "warn_list_footer": "More results are available, use warn list {0} to view them.",
+    "fwarn_list_empty": "No results.",
```
Vgr255: This one though might warrant a more generic place.

src/db.py
```diff
@@ -0,0 +1,807 @@
...
+import os
+import json
+from collections import defaultdict
+
+# increment this whenever making a schema change so that the schema upgrade functions run on start
+# they do not run by default for performance reasons
+SCHEMA_VERSION = 1
+
+def init_vars():
+    with var.GRAVEYARD_LOCK:
```
Vgr255: Shouldn't there be 'conn' in the CM as well?

src/settings.py
```diff
@@ -408,6 +432,17 @@ def plural(role, count=2):
...
+    # converse of plural above (kinda)
+    # this is used to map plural role names back to singular,
+    # so we don't need to worry about stuff like possessives
+    conv = {"wolves": "wolf",
+            "succubi": "succubus"}
+    if plural in conv:
+        return conv[plural]
+    # otherwise we just added an s on the end
+    return plural[:-1]
+
```
Vgr255: What about "wild children" ?

skizzerz: This is only called on team names, will add a comment saying as much

PR #230, opened 2016-07-07
Stop start vote timer if player leaving is the only one who voted

lutoma:
Should fix the bug that just happened in ##werewolf
Untested, might test it tomorrow if nobody else does

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -3251,6 +3251,11 @@ def del_player(cli, nick, forced_death=False, devoice=True, end_game=True, death
                 with var.WARNING_LOCK:
                     var.START_VOTES.discard(nick)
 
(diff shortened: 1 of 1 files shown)


Vgr255:
What happens if there were no votes?

src/wolfgame.py
```diff
@@ -3251,6 +3251,11 @@ def refresh_pl(old_pl):
                 with var.WARNING_LOCK:
                     var.START_VOTES.discard(nick)
 
+                    # Cancel the start vote timer if there are no votes left
+                    if len(var.START_VOTES) < 1 and 'start_votes' in var.TIMERS:
```
skizzerz: use "if not var.START_VOTES" instead, it's cleaner

nyuszika7h:
I don't remember, what was the bug with this?

PR #238, opened 2016-08-01
Add tempban sanction

skizzerz:
Can be either time-based or points-based. Also, made fwarn a bit smarter
at guessing what the user actually wanted to do based on the parameters
given to it. Warnings now always require acknowledgement, because that
paves the way for a future commit only beginning stasis once a warning
is acknowledged.

Warnings also split off into their own file to declutter wolfgame.py a
bit (now only 9k lines, wooooo! >_>)

Does not play nice with eir, that functionality isn't going to be in the
bot itself but rather some custom code in lykos (hooking into privmsg).


src/db.py
```diff
@@ -741,6 +812,10 @@ def _upgrade(oldversion):
                 # player id as a string). When nocasing players, this may cause some records to be merged.
                 with open(os.path.join(dn, "db", "upgrade2.sql"), "rt") as f:
                     c.executescript(f.read())
+            if oldversion < 3:
+                print ("Upgrade from version 2 to 3...", file=sys.stderr)
+                with open(os.path.join(dn, "db", "upgrade3.sql"), "rt") as f:
+                    c.executescript(f.read())
```
Vgr255: Can't you possibly put that in a loop? e.g.:

```
assert oldversion >= 1
for i in range(2, SCHEMA_VERSION+1):
    if oldversion < i:
        print("Upgrading from version {0} to {1}...".format(oldversion, i), file=sys.stderr)
        with open(os.path.join(dn, "db", "upgrade{0}.sql".format(i)), "rt") as f:
            c.executescript(f.read())
        oldversion = i
```

That way you don't have to keep adding identical blocks everytime you perform an update. As for the comments about what each update does, these can go above the loop, e.g. by saying "Upgrade 2 change x, y, and z" and so on.

skizzerz: There is no guarantee that a version upgrade will consist of a simple sql script to execute; there may be python code needed or for things to be broken out in a different way.

Vgr255: Hmm, this possibility doesn't appear to me right now, but you're the DB pro. There are only two upgrade paths right now anyway, so it's not so bad. When we get to 4-5 that are all the same thing I might bring this up again :)

src/db/db.sql
```diff
@@ -159,8 +159,15 @@ CREATE TABLE IF NOT EXISTS access (
...
+CREATE TABLE bantrack (
+	player INTEGER NOT NULL PRIMARY KEY REFERENCES player(id) DEFERRABLE INITIALLY DEFERRED,
+	expires DATETIME,
+	warning_amount INTEGER
+);
+
 -- Used to hold state between restarts
-CREATE TABLE IF NOT EXISTS pre_restart_state (
+CREATE TABLE pre_restart_state (
 	-- List of players to ping after the bot comes back online
```
Vgr255: Hmm, any particular reason why you're unconditionally creating (and presumably overwriting) tables? Sounds like you might outright delete information if you're not careful...

src/wolfgame.py
```diff
@@ -47,6 +46,7 @@
 from src.utilities import *
 from src import db, decorators, events, logger, proxy, debuglog, errlog, plog
 from src.messages import messages
+from src.warnings import *
```
Vgr255: Not too warm on a blanket import here; we do it for utilities but I don't care as much about that ("Just give me the utility functions"). It doesn't matter that much though (especially since you defined `__all__`) so I'm not going to hold up merging if that's the only issue left.

src/db/db.sql
skizzerz: this script is only run on the initial installation; IF NOT EXISTS is superfluous. Also, this will never delete data in any case, it would just error out if the table exists

Vgr255: Well in any case there's no guarantee that the users wouldn't run it themselves. But since this wouldn't delete data, then it's fine.

PR #248, opened 2016-10-25
New IRCContext, Channel and User classes

Vgr255:
Right now these don't do anything, but in the future they will be how
we'll do channel and user handling.

### DIFF
--- a/src/channels.py
+++ b/src/channels.py
@@ -1,48 +1,63 @@
 import time
 
+from enum import Enum
+
 from src.context import IRCContext, Features
 from src.logger import debuglog
 from src import users
 
 Main = None # main channel
 
-all_channels = {}
+_channels = {}
 
 _states = ("not yet joined", "pending join", "joined", "pending leave", "left channel", "", "deleted", "cleared")
 
-def _strip(name):
-    return name.lstrip("".join(Features["STATUSMSG"]))
+class _States(Enum):
+    NotJoined = "not yet joined"
+    PendingJoin = "pending join"
+    Joined = "joined"
+    PendingLeave = "pending leave"
+    Left = "left channel"
 
+    Deleted = "deleted"
+    Cleared = "cleared"
+
+# This is used to tell if this is a fake channel or not. If this
+# function returns a true value, then it's a fake channel. This is
+# useful for testing, where we might want the users in fake channels.
 def predicate(name):
     return not name.startswith(tuple(Features["CHANTYPES"]))
 
-def get(name):
-    """Return an existing channel, or raise a KeyError if it doesn't exist."""
+get = _channels.__getitem__
 
-    return all_channels[_strip(name)]
-
-def add(name, cli):
+def add(name, cli, key=""):
     """Add and return a new channel, or an existing one if it exists."""
 
-    name = _strip(name)
+    # We use add() in a bunch of places where the channel probably (but
+    # not surely) already exists. If it does, obviously we want to use
+    # that one. However, if the client is *not* the same, that means we
+    # would be trying to send something from one connection over to
+    # another one (or some other weird stuff like that). Instead of
+    # jumping through hoops, we just disallow it here.
 
-    if name in all_channels:
-        if cli is not all_channels[name].client:
+    if name in _channels:
+        if cli is not _channels[name].client:
             raise RuntimeError("different IRC client for channel {0}".format(name))
-        return all_channels[name]
+        return _channels[name]
 
     cls = Channel
     if predicate(name):
         cls = FakeChannel
 
-    chan = all_channels[name] = cls(name, cli)
-    chan.join()
+    chan = _channels[name] = cls(name, cli)
+    chan.join(key)
     return chan
 
-def exists(name):
-    """Return True if a channel with the name exists, False otherwise."""
+exists = _channels.__contains__
 
-    return _strip(name) in all_channels
+def channels():
+    """Iterate over all the current channels."""
+    yield from _channels.values()
 
 class Channel(IRCContext):
 
@@ -53,37 +68,51 @@ class Channel(IRCContext):
         self.users = set()
         self.modes = {}
         self.timestamp = None
-        self.state = 0
+        self.state = _States.NotJoined
 
     def __del__(self):
         self.users.clear()
         self.modes.clear()
-        self.state = -2
+        self.state = _States.Deleted
         self.client = None
         self.timestamp = None
 
     def __str__(self):
-        return "{self.__class__.__name__}: {self.name} ({0})".format(_states[self.state], self=self)
+        return "{self.__class__.__name__}: {self.name} ({0})".format(self.state.value, self=self)
 
     def __repr__(self):
         return "{self.__class__.__name__}({self.name!r})".format(self=self)
 
     def join(self, key=""):
-        if self.state in (0, 4):
-            self.state = 1
+        if self.state in (_States.NotJoined, _States.Left):
+            self.state = _States.PendingJoin
             self.client.send("JOIN {0} :{1}".format(self.name, key))
 
     def part(self, message=""):
-        if self.state == 2:
-            self.state = 3
+        if self.state is _States.Joined:
+            self.state = _States.PendingLeave
             self.client.send("PART {0} :{1}".format(self.name, message))
 
     def kick(self, target, message=""):
-        if self.state == 2:
+        if self.state is _States.Joined:
             self.client.send("KICK {0} {1} :{2}".format(self.name, target, message))
 
     def mode(self, *changes):
-        if not changes:
+        """Perform a mode change on the channel.
+
+        Usage:
+
+            chan.mode() # Will get back the modes on the channel
+            chan.mode("b") # Will get the banlist back
+            chan.mode("-m")
+            chan.mode(["-v", "woffle"], ["+o", "Vgr"])
+            chan.mode("-m", ("+v", "jacob1"), ("-o", "nyuszika7h"))
+
+        This performs both single and complex mode changes.
+
+        """
+
+        if not changes: # bare call; get channel modes
             self.client.send("MODE", self.name)
             return
 
@@ -93,7 +122,7 @@ class Channel(IRCContext):
             if isinstance(change, str):
                 change = (change, None)
             params.append(change)
-        params.sort(key=lambda x: x[0][0])
+        params.sort(key=lambda x: x[0][0]) # sort by prefix
 
         while params:
             cur, params = params[:max_modes], params[max_modes:]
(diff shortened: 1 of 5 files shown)


src/context.py
```diff
@@ -0,0 +1,69 @@
+Features = {"CASEMAPPING": "rfc1459", "CHARSET": "utf-8", "STATUSMSG": {"@", "+"}, "CHANTYPES": {"#"}} # IRC server features (these are defaults)
```
skizzerz: This comment doesn't need to exist (it was obvious just by looking at the line). The value for CASEMAPPING and CHARSET should both be enums

src/context.py
```diff
@@ -0,0 +1,69 @@
...
+        # 2 (1) - The space between the target and the data
+        # 3 (1) - The colon at the front of the data to send
+        # 4 (3) - I don't know why, but we need 3 more/less characters
+        length = 0x200 - 7
+        # Next, we need to reduce the length to account for our address
+        length -= len(full_address)
+        # Then we also need to account for the target's length
+        length -= len(name)
+        # Finally, we need to account for the send type's length
+        length -= len(send_type)
```
skizzerz: This is incredibly flawed. You aren't including full_address in the sent message so there is no need to deduct it. Remember also that the trailing \r\n is included in the 512 byte count. Don't be clever with 0x200, it's less obvious than 512 -- just use that.

Your comment has redundant lines and incorrect lines, clean it up

src/context.py
```diff
@@ -0,0 +1,69 @@
...
+
+        for line in data.splitlines():
+            while line:
+                extra, line = line[:length], line[length:]
```
skizzerz: this is terrible, you should instead adapt the code from break_long_message which ensures that lines are split on word boundaries instead of in the middle of words. I think it also correctly handles things like re-applying bold/color codes as needed.

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+    - allow_none (defaulting to False) allows no match at all, and
+      returns None instead of raising an error; an empty list will be
+      returned if this is used with allow_multiple;
+
+    - allow_bot (defaulting to False) allows the bot to be matched and
+      returned;
+
+    - raw_nick (defaulting to False) means that the nick has not been
+      yet parsed, and so ident and host will be None, and nick will be
+      a raw nick of the form nick!ident@host.
```
skizzerz: Make raw_nick a separate function

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+    """Create a new user, add it to the user list and return it.
+
+    This function takes up to 6 keyword-only arguments (and no positional
+    arguments): nick, ident, host, realname, account and channels.
+    With the exception of the first one, any parameter can be omitted.
+    If a matching user already exists, a ValueError will be raised.
+
+    The raw_nick keyword argument may be set if the nick has not yet
+    been parsed. In that case, ident and host must both be None, and
+    nick must be in the form nick!ident@host.
```
skizzerz: again, make raw_nick handling a different function

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+    """Return True if a matching user exists.
+
+    Positional and keyword arguments are the same as get(), with the
+    exception that allow_none may not be used (a RuntimeError will be
+    raised in that case).
+
+    """
+
+    if allow_none: # why would you even want to do that?
+        raise RuntimeError("Cannot use allow_none=True with exists()")
```
skizzerz: pointless, just forward everything as-is and return False if None is returned.

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+
+_raw_nick_pattern = re.compile(
+
+    r"""
+    \A
+    (?P<nick>  [^!@\s]+ (?=!|$) )? !?
+    (?P<ident> [^!@\s]+         )? @?
+    (?P<host>  \S+ )?
+    \Z
+    """,
```
skizzerz: Overly complicated, just use non-greedy matching for the capture groups instead of this mess. Prefer shorter regex over splitting into multiple lines, this way manages to make it LESS readable

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def __repr__(self):
+        return "{self.__class__.__name__}({self.nick!r}, {self.ident!r}, {self.host!r}, {self.realname!r}, {self.account!r}, {self.channels!r})".format(self=self)
+
+    def lower(self):
+        return type(self)(self.client, lower(self.nick), lower(self.ident), lower(self.host), lower(self.realname), lower(self.account), channels, ref=(self.ref or self))
+
+    def is_owner(self):
+        if self.is_fake:
+            return False # fake nicks can't ever be owner
```
skizzerz: pointless comment (as is the one for is_admin)

src/users.py
```diff
@@ -0,0 +1,440 @@
...
+    def stasis_count(self):
+        """Return the number of games the user is in stasis for."""
+        temp = self.lower()
+        amount = 0
+
+        if not var.DISABLE_ACCOUNTS:
+            amount = var.STASISED_ACCS.get(temp.account, 0)
+
+        for hostmask in var.STASISED:
+            if temp.match_hostmask(hostmask):
```
skizzerz: wildcard hostmasks are not supported for stasis, just do a direct lookup

src/context.py
jacob1: It isn't that flawed, since full_address actually does need to be subtracted (due to the fact that other users will get full_address when they receive the message). But the 0x200 should be changed and the "I don't know why" comment should be actually explained. The 3 missing characters are a colon at the beginning before the hostmask, and the \r\n at the end.

It looks like this:
:hostmask PRIVMSG channel :message\r\n

nyuszika7h:
Test please ignore

src/users.py
Vgr255: If it's pointless, then I'd rather just do nothing with it. Passing allow_none=True to exists() is downright stupid and I have no sympathy for you if you do that.

src/users.py
Vgr255: I hope you're not talking about the regex in match_hostmask, because it's impossible to reason about. Having it on multiple lines makes it a lot easier to reason about. I'll do non-greedy though, forgot about that.

nyuszika7h:
test please ignore

src/context.py
skizzerz: We only handle very specific values of this, so we should have an Enum encoding the values we support. If the ircd sends an unsupported value, we'll just default it.

src/context.py
skizzerz: This is a public function. If it's meant to be the lowest level then it should be private. Furthermore, I'd argue that we should raise instead of splitting the message here in that case, as it means the caller did not properly call break_long_message before calling us

src/users.py
skizzerz: It makes for a much cleaner and safer API to just enforce blah(hostmask) or blah(nick, ident, host) (no default value for ident/host so that you must pass all 3 if calling that overload). Since python doesn't have function overloading of that form (at least not in 3.4), they need to be separate function names instead.

src/users.py
skizzerz: this isn't doing nothing, this is adding pointless code that makes no difference at all. Doing nothing would be exactly what I stated in my review, and I still stand by that statement. Passing stupid things to a function is better handled at review time

src/users.py
skizzerz: The more fancy characters you introduce into a regex, the harder it is to reason about. Breaking it into multiple lines also makes it more difficult to see the shape of the string it is matching. Someone familiar with regular expressions should be able to look at a short regex (<20 characters) and pretty much instantly see what it is doing. I have no desire to make our regular expressions "understandable" by people not already familiar with them, that is what comments are for

src/users.py
skizzerz: Actually, my main issue with this function is that it takes *args and *kwargs to begin with. I'd rather this just duplicate the signature of get() minus the specific things you don't wish to support (in other words, you don't have an allow_none kwarg).

src/users.py
Vgr255: Actually, it does, and it's called `functools.singledispatch`.

src/context.py
Vgr255: This function does what `IRCClient.msg` did before. The only difference is that it allows more characters instead of stupidly hardcoding the max. I'm also not raising anything from that one function, because I'd much prefer having a very minor annoyance (message being split in two) which *doesn't* impact the gameplay or anything of the sort, than an error which then prevents anything else from happening. So no, it's better to split like it's done right now and handle proper message splitting further up. If you're *that* concerned, we can consider updating `.send()` instead - but `.raw_send()` is staying as-is.

src/channels.py
```diff
@@ -0,0 +1,223 @@
...
+    def __str__(self):
+        return "{self.__class__.__name__}: {self.name} ({0})".format(_states[self.state], self=self)
+
+    def __repr__(self):
+        return "{self.__class__.__name__}({self.name!r})".format(self=self)
+
+    def join(self, key=""):
+        if self.state in (0, 4):
+            self.state = 1
+            self.client.send("JOIN {0} :{1}".format(self.name, key))
```
Vgr255: Every command which may take more than one argument takes a colon. Having nothing after the colon means that there's no key. It's perfectly valid as per the spec.

src/users.py
Vgr255: Ok for comments, but there isn't much to actually simplify in this regex.

src/channels.py
skizzerz: Incorrect. Only commands where the argument may contain spaces takes a colon. JOIN is not one of those cases, in fact it's the opposite. JOIN #channel1,#channel2 key1 key2 is the proper syntax. Re-read RFC 1459 if you don't believe me, but JOIN does **not** have a colon in it.

src/users.py
skizzerz: See IRC for why that is idiotic reasoning, I don't feel like repeating it here. This is not something I'm willing to compromise on -- this PR will not be merged unless the signature is fixed

src/channels.py
skizzerz: Yeah, I'll retract this comment, it's fine as-is (I was reading the spec wrong)

src/users.py
Vgr255: I'll do that, but in the meantime I suggest you bring down your tone a notch. However right or wrong either of us may be, that kind of reply isn't pleasant.

src/context.py
skizzerz: Ok, sounds good to me. Only remaining thing is whether or not you want this to be public -- if you do then it's fine as-is, if you don't I'd suggest renaming to _send. My personal preference would be to make it private (assuming that works with inheritance), to make it clear that it should only be called from something that implements the class rather than outside callers. I don't think we want to encourage people to be bypassing our send stuff (and if they really need to for some reason, python lets them do it anyway)

src/users.py
skizzerz: You haven't really replied on IRC so I'll just copy it here for posterity. Sorry about the tone, I've been having a bad week (don't want to go into it), but I shouldn't be taking it out on you.

```
<woffle> Vgr: and since I didn't really delve into why on the review, APIs that can take any arbitrary arguments even though it only cares about some of them are terrible for developers to use. It kills any sort of editor help/completion you may have available, and it forces you to look in another completely separate spot to see how to actually use it, since the function signature itself no longer tells you
<woffle> If you're taking in *args or **kwargs, that better be because you actually do allow any arbitrary arguments, not because you're being lazy
<woffle> that exact same reason is why I killed your other change to events
<woffle> s/killed/re-did/
```

src/context.py
Vgr255: (Note for whoever seeing this not-out-of-date review: `raw_send()` was renamed to `_send()`)

src/channels.py
```diff
@@ -0,0 +1,223 @@
...
+                final.append(mode)
+
+            for target in targets:
+                if target is not None:
+                    final.append(" ")
+                    final.append(target)
+
+            self.client.send("MODE", self.name, "".join(final))
+
+    def update_modes(self, rawnick, mode, targets):
```
Vgr255: Docstring added.

src/context.py
Vgr255: Proper length calculation and comments updated.

PR #251, opened 2016-10-28
Change the user handling calls in wolfgame.py

Vgr255:
Nothing more or less than a change in the interface

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -46,18 +46,13 @@ import botconfig
 import src
 import src.settings as var
 from src.utilities import *
-from src import db, decorators, events, logger, proxy, debuglog, errlog, plog
+from src import db, decorators, events, users, logger, proxy, debuglog, errlog, plog
+from src.decorators import cmd, hook, handle_error, event_listener, COMMANDS
 from src.messages import messages
 from src.warnings import *
 
 # done this way so that events is accessible in !eval (useful for debugging)
 Event = events.Event
-
-cmd = decorators.cmd
-hook = decorators.hook
-handle_error = decorators.handle_error
-event_listener = decorators.event_listener
-COMMANDS = decorators.COMMANDS
 
 # Game Logic Begins:
 
@@ -176,7 +171,7 @@ def connect_callback(cli):
         var.DISABLE_ACCOUNTS = True
         var.ACCOUNTS_ONLY = False
 
-        if nick in var.USERS:
+        if users.exists(nick, user, host):
             return
 
         if nick == botconfig.NICK:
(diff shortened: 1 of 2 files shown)


src/wolfgame.py
```diff
@@ -7759,3 +7754,4 @@ def frole(cli, nick, chan, rest):
             del COMMANDS[comd]
 
 # vim: set sw=4 expandtab:
+
```
ilbelkyr: Why the blank line?

src/wolfgame.py
```diff
@@ -7432,15 +7432,15 @@ def pyeval(cli, nick, chan, rest):
             else:
                 cli.msg(chan, a[:500])
         except Exception as e:
-            cli.msg(chan, str(type(e))+":"+str(e))
+            cli.msg(chan, repr(e))
```
ilbelkyr: hm I might be missing something but how does this relate to the PR?

src/wolfgame.py
Vgr255: No clue. I know I didn't manually put it there, so my guess is funny line ending shenanigans.

src/wolfgame.py
Vgr255: It doesn't. I just did it alongside when I was testing. I figured it was so minor that it didn't matter, but if it does I'll just push it directly.

src/wolfgame.py
```diff
@@ -263,7 +259,7 @@ def on_banlist_end(cli, svr, nick, chan, *etc):
     else:
         cli.who(botconfig.CHANNEL, "%uhsnfa")
 
-@hook("mode")
+@hook("mode") # not bothering to update this function - it was already re-implemented in channels.Channel.update_modes
```
skizzerz: if this function is obsolete by the refactor, mark with a TODO or FIXME comment indicating it should be removed at a later date. Otherwise, remove the comment entirely; code comments are for explaining why the code is doing what it is doing or how to do certain things, not for general remarks about a process you are performing. Reminders for later using a tag like TODO or FIXME allows us to easily scan the repo for such tags to ensure that they eventually get done.

src/wolfgame.py
```diff
@@ -1267,7 +1263,7 @@ def fjoin(cli, nick, chan, rest):
                 continue
         if not tojoin:
             continue
-        ul = list(var.USERS.keys())
+        ul = list(var.USERS.keys()) # ewww, not touching that! users.User.lower() allows us to do that *much* better
```
skizzerz: Pointless comment, remove

src/wolfgame.py
```diff
@@ -3305,7 +3301,7 @@ def goat(cli, nick, chan, rest):
         cli.notice(nick, messages["goat_fail"])
         return
 
-    ul = list(var.USERS.keys())
+    ul = list(var.USERS.keys()) # ewwww²
```
skizzerz: Pointless comment, remove

src/wolfgame.py
```diff
@@ -3514,7 +3509,7 @@ def rename_player(cli, prefix, nick):
         var.NO_LYNCH.remove(prefix)
         var.NO_LYNCH.add(nick)
 
-@hook("nick")
+@hook("nick") # Not updating this function - it'll need a complete overhaul anyway
```
skizzerz: Another likely candidate for a TODO/FIXME tag

PR #259, opened 2016-11-04
Fflags fix, fixes #254

### LINKED ISSUE #254: !fflags overwrites old flags
It seems if I do e.g. `!fflags moo +j`, and they had `+a`, it's removed, even though it's not supposed to, only with `=`.

jacob1:
Fixes to #254 and also another issue we've had before

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6421,7 +6421,7 @@ def fflags(cli, nick, chan, rest):
     else:
         acc, hm = parse_warning_target(params[0])
         flags = params[1]
-        cur_flags = set(var.FLAGS_ACCS[acc] + var.FLAGS[hm])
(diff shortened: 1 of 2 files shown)


src/wolfgame.py
```diff
@@ -6421,7 +6421,7 @@ def fflags(cli, nick, chan, rest):
     else:
         acc, hm = parse_warning_target(params[0])
         flags = params[1]
-        cur_flags = set(var.FLAGS_ACCS[acc] + var.FLAGS[hm])
+        cur_flags = set(var.FLAGS_ACCS[acc.lower()] + var.FLAGS[hm.lower()])
```
skizzerz: account names should be lowercased via irc_lower() so they casefold properly. Hostmask is good as-is.

PR #262, opened 2016-11-04
Disallow pingif greater than or equal to var.MAX_PLAYERS

nyuszika7h:
There is no good reason to set your ping preferences to equal or exceed `var.MAX_PLAYERS`, as you won't be able to join at that point anyway, not even admins can `!fjoin` you.


Vgr255:
Allowing `pingif` to select a higher number than the max number of players was a deliberate choice. The max number of players may change depending on the game mode, `fgame roles` increases it, and in the future it may change.

PR #272, opened 2016-11-19
Add a transition method for the old->new interface

Vgr255:
The argument should be `True` for all commands until the new interface is implemented, at which point I'll begin updating the calls. Until then, there's absolutely no change in functionality whatsoever.

(I checked that all commands had it with a simple `assert old_api` in `cmd.__init__`)

### DIFF
--- a/src/decorators.py
+++ b/src/decorators.py
@@ -223,7 +223,13 @@ class cmd:
         return self
 
     @handle_error
-    def caller(self, *args):
+    def caller(self, var, wrapper, message):
+        # The wrapper is an object which will know the sender and target
+        # It will have methods such as .reply(), taking off the load from the end code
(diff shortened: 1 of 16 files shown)


skizzerz:
Reverted in 11d626ab25a8bb690962036e4a2b0a913feabdb7 for causing too many issues

A better way of going about this would be to give the new API a different name, such as by using command instead of cmd. Then, when ready to drop the backwards compat layer, we can delete cmd() and python will alert us whenever a nonexistent function gets called, instead of giving a traceback at a useless point in time. Still easy to see/grep for things that didn't get updated (if you get hits for cmd, it isn't updated yet) so easy for developers, and easier for bot admins too because the backwards compat is on by default instead of off by default

PR #274, opened 2016-11-21
Add a backwards-compatibility handler for commands

Vgr255:
It's actually ready to merge now, and it addresses @skizzerz's concerns.


skizzerz:
Will give it a more thorough look later to ensure call sites were not missed, but I'm still of the opinion that this should not be merged until the new API is at least present (even if it doesn't fully work).

Vgr255:
Sure, it doesn't need to be merged right now. I'll push the new API when it's done to this branch, so this will be a slightly bigger PR to review if both get merged at once. But then you can review what's already here (very unlikely to change), and review the new commit when I do push it.

PR #277, opened 2016-12-19
Improve user part/quit/nick hooks

Vgr255:
Tested thanks to mark-otaris, should be working fine.


src/wolfgame.py
```diff
@@ -3284,35 +3244,29 @@ def fgoat(cli, nick, chan, rest):
...
+                        roleset.remove("(dced)" + temp.nick)
+                        roleset.add(temp.nick)
+
+                if temp.nick in var.DCED_PLAYERS:
+                    var.PLAYERS[temp.nick] = var.DCED_PLAYERS.pop(temp.nick)
+
                 if show_message:
-                    cli.mode(chan, "+v", nick, nick+"!*@*")
-                    cli.msg(chan, messages["player_return"].format(nick))
+                    channels.Main.mode(("+" + hooks.Features["PREFIX"]["+"], target))
```
nyuszika7h: Why don't we have some `voice()` function?

Vgr255: I was thinking the same thing kinda, but we don't do mode changes that much. And you'd need two methods, one for voicing and one for devoicing, or you use a keyword-only argument, and in such cases you can't handle voicing and devoicing in a single change. I prefer that, where it makes sense to, just use .mode instead. If I do add the method though, it won't be as part of the PR (either before or after it's merged).

src/users.py
```diff
@@ -157,6 +157,21 @@ def __iter__(self):
     def items():
         yield from var.USERS.items()
 
+def complete_match(string, users):
+    matches = []
+    string = lower(string)
```
skizzerz: we know we're matching nicks here, so probably should use irc_lower()

Vgr255: `lower` *is* `irc_lower`, except better. The latter is going away after the transition.

Vgr255: As an aside (and maybe I wasn't clear on that part), I'm trying to kill off `src.utilities`. That doesn't mean to kill all functions from it (although a great chunk of them will get killed off), but it does mean moving some things elsewhere. It may not be entirely possible right away, but right now none of the new interface relies on it. Let's not make our job harder by relying on it ;)

Vgr255:
Can this be merged? I'm waiting on this to work on other things.

Issue #176, opened 2015-10-17
Use a more clear message for ambiguous game modes


nyuszika7h:
It might be nice if the bot said something like this:

```
-lykos- Ambiguous game mode: al. Possible game modes: aleatoire, alpha
```

Fudster:
Which way would you prefer to do it then?

- list all game modes.
- list all *possible* game modes.

nyuszika7h:
If it's a partial match, it should only list the possible ones. If there's no match, it *could* possibly list all game modes, or just tell you to use `!modes`.

PR #279, opened 2017-01-18
Fix for #176 improve invalid role errors

Fudster:
Issue: #176 
I tweaked complete_match and added complete_one_match, along with message invalid_mode.


skizzerz:
PR should be based on master, not stable

nyuszika7h:
Just noticed this. Next time you can edit and use "Change base" instead of having to open a new PR.

PR #280, opened 2017-01-18
Fix for #176 improve invalid role errors

Fudster:
Issue: #176
I tweaked complete_match and added complete_one_match, along with message invalid_mode.

### DIFF
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -417,7 +417,7 @@ def complete_match(string, matches):
             return [string]
         if possible.startswith(string) or possible.lstrip("[{\\^_`|}]").startswith(string):
             possible_matches.add(possible)
-    return sorted(possible_matches))
+    return sorted(possible_matches)
 
 def complete_one_match(string, matches):
     match = complete_match(string,matches) 
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -417,7 +417,7 @@ def complete_match(string, matches):
             return [string]
         if possible.startswith(string) or possible.lstrip("[{\\^_`|}]").startswith(string):
             possible_matches.add(possible)
-    return sorted(list(possible_matches))
+    return sorted(possible_matches))
 
 def complete_one_match(string, matches):
     match = complete_match(string,matches) 
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -15,7 +15,7 @@ __all__ = ["pm", "is_fake_nick", "mass_mode", "mass_privmsg", "reply",
            "is_owner", "is_admin", "plural", "singular", "list_players",
            "list_players_and_roles", "list_participants", "get_role", "get_roles",
            "get_reveal_role", "get_templates", "role_order", "break_long_message",
-           "complete_match", "get_victim", "get_nick", "InvalidModeException"]
+           "complete_match","complete_one_match", "get_victim", "get_nick", "InvalidModeException"]
 # message either privmsg or notice, depending on user settings
 def pm(cli, target, message):
     if is_fake_nick(target) and botconfig.DEBUG_MODE:
@@ -411,18 +411,20 @@ def break_long_message(phrases, joinstr = " "):
 
 #completes a partial nickname or string from a list
 def complete_match(string, matches):
-    num_matches = 0
-    bestmatch = string
+    possible_matches = set()
     for possible in matches:
         if string == possible:
-            return string, 1
+            return {string}
         if possible.startswith(string) or possible.lstrip("[{\\^_`|}]").startswith(string):
-            bestmatch = possible
-            num_matches += 1
-    if num_matches != 1:
-        return None, num_matches
-    else:
-        return bestmatch, 1
+            possible_matches.add(possible)
+    return sorted(list(possible_matches))
+
+#wrapper around complete_match() to return a single string
+def complete_one_match(string, matches):
+    match = complete_match(string,matches) 
+    if len(match) == 1:
+        return match.pop()
+    return None
 
 #wrapper around complete_match() used for roles
 def get_victim(cli, nick, victim, in_chan, self_in_list=False, bot_in_list=False):
@@ -437,20 +439,20 @@ def get_victim(cli, nick, victim, in_chan, self_in_list=False, bot_in_list=False
         pl.append(botconfig.NICK)
         pll.append(botconfig.NICK.lower())
 
-    tempvictim, num_matches = complete_match(victim.lower(), pll)
-    if not tempvictim:
+    tempvictims = complete_match(victim.lower(), pll)
+    if len(tempvictims) != 1:
         #ensure messages about not being able to act on yourself work
-        if num_matches == 0 and nick.lower().startswith(victim.lower()):
+        if len(tempvictims) == 0 and nick.lower().startswith(victim.lower()):
             return nick
         reply(cli, nick, chan, messages["not_playing"].format(victim), private=True)
         return
-    return pl[pll.index(tempvictim)] #convert back to normal casing
+    return pl[pll.index(tempvictims.pop())] #convert back to normal casing
 
 # wrapper around complete_match() used for any nick on the channel
 def get_nick(cli, nick):
     ul = [x for x in var.USERS]
     ull = [x.lower() for x in var.USERS]
-    lnick, num_matches = complete_match(nick.lower(), ull)
+    lnick = complete_match(nick.lower(), ull)
     if not lnick:
         return None
     return ul[ull.index(lnick)]
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -420,9 +420,9 @@ def complete_match(string, matches):
     return sorted(possible_matches)
 
 def complete_one_match(string, matches):
-    match = complete_match(string,matches) 
-    if len(match) == 1:
-        return match.pop()
+    matches = complete_match(string,matches) 
+    if len(matches) == 1:
+        return matches.pop()
     return None
 
 #wrapper around complete_match() used for roles
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -414,12 +414,11 @@ def complete_match(string, matches):
     possible_matches = set()
     for possible in matches:
         if string == possible:
-            return {string}
+            return [string]
         if possible.startswith(string) or possible.lstrip("[{\\^_`|}]").startswith(string):
             possible_matches.add(possible)
     return sorted(list(possible_matches))
 
-#wrapper around complete_match() to return a single string
 def complete_one_match(string, matches):
     match = complete_match(string,matches) 
     if len(match) == 1:
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6707,7 +6707,7 @@ def listroles(cli, nick, chan, rest):
     elif rest[0] and not rest[0].isdigit():
         gamemode = rest[0]
         if gamemode not in var.GAME_MODES.keys():
-            matches = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
+            matches = complete_match(rest[0], var.GAME_MODES.keys() - {"roles", "villagergame"} - var.DISABLED_GAMEMODES)
             if len(matches) > 1:
                 reply(cli, nick, chan, nick + " " + messages["invalid_mode"].format(rest[0], ", ".join(matches)))
                 rest = []
@@ -7014,7 +7014,7 @@ def vote_gamemode(var, wrapper, gamemode, doreply):
         return
 
     if gamemode not in var.GAME_MODES.keys():
-        matches = complete_match(gamemode, var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
+        matches = complete_match(gamemode, var.GAME_MODES.keys() - {"roles", "villagergame"} - var.DISABLED_GAMEMODES)
         if not matches:
             wrapper.pm(messages["invalid_mode_no_list"].format(gamemode))
             return
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6708,29 +6708,35 @@ def listroles(cli, nick, chan, rest):
         gamemode = rest[0]
         if gamemode not in var.GAME_MODES.keys():
             matches = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
-            if len(gamemode) > 1:
-                msg.append(messages["invalid_mode"].format(nick, rest[0], ", ".join(matches)))
+            if len(matches) > 1:
+                reply(cli, nick, chan, nick + " " + messages["invalid_mode"].format(rest[0], ", ".join(matches)))
                 rest = []
                 roleindex = {}
+                return
+            if len(matches) == 0:
+                reply(cli, nick, chan, nick + " " + messages["invalid_mode_no_list"].format(rest[0], ", ".join(matches)))
+                rest = []
+                roleindex = {}
+                return
             else:
-                gamemode = matches.pop()
+                gamemode = matches[0]
+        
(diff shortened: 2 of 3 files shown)


src/utilities.py
```diff
@@ -411,18 +411,20 @@ def break_long_message(phrases, joinstr = " "):
...
+            return {string}
         if possible.startswith(string) or possible.lstrip("[{\\^_`|}]").startswith(string):
-            bestmatch = possible
-            num_matches += 1
-    if num_matches != 1:
-        return None, num_matches
-    else:
-        return bestmatch, 1
+            possible_matches.add(possible)
+    return sorted(list(possible_matches))
```
skizzerz: mismatched return types - set and list

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
...
+                gamemode = gamemode.pop()
+        else:
+            validMode = gamemode in var.GAME_MODES.keys() and gamemode != "roles" and gamemode != "villagergame" and gamemode not in var.DISABLED_GAMEMODES
+            if validMode and hasattr(var.GAME_MODES[gamemode][0](), "ROLE_GUIDE"):
+                mode = var.GAME_MODES[gamemode][0]()
             if hasattr(mode, "ROLE_INDEX") and hasattr(mode, "ROLE_GUIDE"):
                 roleindex = mode.ROLE_INDEX
                 roleguide = mode.ROLE_GUIDE
-            elif gamemode == "default" and "ROLE_INDEX" in var.ORIGINAL_SETTINGS and "ROLE_GUIDE" in var.ORIGINAL_SETTINGS:
+            elif gamemode == "default" and "ROLE_INDEX" in var.ORIGINAL_SETTINGS and "ROLE_GUIDE" in var.   ORIGINAL_SETTINGS:
```
skizzerz: spurious whitespace

src/utilities.py
```diff
@@ -411,18 +411,20 @@ def break_long_message(phrases, joinstr = " "):
...
+
+#wrapper around complete_match() to return a single string
```
skizzerz: pointless comment.
1. Comments about what a function does should be in docstrings
2. If the function is obvious (as this one is), no comment is needed

src/utilities.py
```diff
@@ -411,18 +411,20 @@ def break_long_message(phrases, joinstr = " "):
...
+def complete_one_match(string, matches):
+    match = complete_match(string,matches)
```
skizzerz: missing space after comma

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
     elif rest[0] and not rest[0].isdigit():
         gamemode = rest[0]
         if gamemode not in var.GAME_MODES.keys():
-            gamemode, _ = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
-        validMode = gamemode in var.GAME_MODES.keys() and gamemode != "roles" and gamemode != "villagergame" and gamemode not in var.DISABLED_GAMEMODES
-        if validMode and hasattr(var.GAME_MODES[gamemode][0](), "ROLE_GUIDE"):
-            mode = var.GAME_MODES[gamemode][0]()
+            gamemode = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
```
skizzerz: don't smash the gamemode variable here, pick a different variable name

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
...
+            if len(gamemode) > 1:
+                msg.append("{0}: {1} is not a valid game mode. Possbile modes are: {2} ".format(nick, rest[0], ", ".join(gamemode)))
```
skizzerz: No hardcoded strings, messages must be added to en.json and referred to by key. Also check your spelling

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
...
+                rest = []
+                roleindex = {}
+            else:
+                gamemode = gamemode.pop()
+        else:
+            validMode = gamemode in var.GAME_MODES.keys() and gamemode != "roles" and gamemode != "villagergame" and gamemode not in var.DISABLED_GAMEMODES
```
skizzerz: this is fugly, use set union to check for disabled gamemodes, roles, and villagergame in one operation. I know this wasn't you but since you're touching the code anyway, you might as well fix this up

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
...
             else:
-                msg.append("{0}: {1} is not a valid game mode.".format(nick, rest[0]))
-            rest = []
-            roleindex = {}
+                if validMode and not hasattr(var.GAME_MODES[gamemode][0](), "ROLE_GUIDE"):
+                    msg.append("{0}: {1}roles is disabled for the {2} game mode.".format(nick, botconfig.CMD_CHAR, gamemode))
+                else:
+                    msg.append("{0}: {1} is not a valid game mode.".format(nick, rest[0]))
+                    rest = []
+                    roleindex = {}
```
skizzerz: these 2 lines are further indented compared to previous. Is this intentional and if so, why?

src/wolfgame.py
```diff
@@ -6902,10 +6908,12 @@ def game_stats(cli, nick, chan, rest):
     if len(rest) and not rest[0].isdigit():
         gamemode = rest[0]
         if gamemode != "all" and gamemode not in var.GAME_MODES.keys():
-            gamemode, _ = complete_match(gamemode, var.GAME_MODES.keys())
-        if not gamemode:
-            cli.notice(nick, messages["invalid_mode_no_list"].format(rest[0]))
-            return
+            gamemode = complete_match(gamemode, var.GAME_MODES.keys())
```
skizzerz: again, use a different variable name. A list isn't a gamemode, so don't call it that

src/wolfgame.py
```diff
@@ -6972,14 +6980,17 @@ def player_stats(cli, nick, chan, rest):
...
                 reply(cli, nick, chan, messages["no_such_role"].format(role))
                 return
-            role = match
+            if len(match) > 2:
+                reply(cli, nick, chan, messages["invalid_mode"].format(role, ", ".join(match)))
+                return    
         # Attempt to find the player's stats
         reply(cli, nick, chan, db.get_player_stats(acc, hostmask, role))
 
+
```
skizzerz: excess whitespace

src/wolfgame.py
```diff
@@ -6994,13 +7005,16 @@ def vote_gamemode(var, wrapper, gamemode, doreply):
         return
 
     if gamemode not in var.GAME_MODES.keys():
-        match, _ = complete_match(gamemode, var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
+        match = complete_match(gamemode, var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
```
skizzerz: bad variable name, a list isn't singular

src/utilities.py
Fudster: I switched the statement above it to a list.

Using lists here so the matches would be in order.

Vgr255: Using `sorted(list(x))` is pointless. Just use `sorted(x)`.

src/utilities.py
```diff
@@ -411,18 +411,19 @@ def break_long_message(phrases, joinstr = " "):
...
-            num_matches += 1
-    if num_matches != 1:
-        return None, num_matches
-    else:
-        return bestmatch, 1
+            possible_matches.add(possible)
+    return sorted(possible_matches)
+
+def complete_one_match(string, matches):
+    match = complete_match(string,matches)
```
Vgr255: Bad variable name - make it `matches` since there may be more than one.

src/utilities.py
```diff
@@ -411,18 +411,19 @@ def break_long_message(phrases, joinstr = " "):
...
-    else:
-        return bestmatch, 1
+            possible_matches.add(possible)
+    return sorted(possible_matches)
+
+def complete_one_match(string, matches):
+    match = complete_match(string,matches) 
+    if len(match) == 1:
+        return match.pop()
+    return None
```
Vgr255: Not needed in practice (Python adds an implicit `return None` at the end of every function), but you could keep it as a "Returning None is intended" hint.

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
     elif rest[0] and not rest[0].isdigit():
         gamemode = rest[0]
         if gamemode not in var.GAME_MODES.keys():
-            gamemode, _ = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
-        validMode = gamemode in var.GAME_MODES.keys() and gamemode != "roles" and gamemode != "villagergame" and gamemode not in var.DISABLED_GAMEMODES
-        if validMode and hasattr(var.GAME_MODES[gamemode][0](), "ROLE_GUIDE"):
-            mode = var.GAME_MODES[gamemode][0]()
+            matches = complete_match(rest[0], var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
+            if len(gamemode) > 1:
```
Vgr255: You want `matches`, not `gamemode`, here.

src/wolfgame.py
```diff
@@ -6707,24 +6707,30 @@ def listroles(cli, nick, chan, rest):
...
+                msg.append(messages["invalid_mode"].format(nick, rest[0], ", ".join(matches)))
+                rest = []
+                roleindex = {}
+            else:
+                gamemode = matches.pop()
+        else:
```
Vgr255: What is the purpose of this else clause?

src/wolfgame.py
```diff
@@ -6902,10 +6908,12 @@ def game_stats(cli, nick, chan, rest):
     if len(rest) and not rest[0].isdigit():
         gamemode = rest[0]
         if gamemode != "all" and gamemode not in var.GAME_MODES.keys():
-            gamemode, _ = complete_match(gamemode, var.GAME_MODES.keys())
-        if not gamemode:
-            cli.notice(nick, messages["invalid_mode_no_list"].format(rest[0]))
-            return
+            matches = complete_match(gamemode, var.GAME_MODES.keys())
+            if len(gamemode) == 1:
```
Vgr255: Still wrong variable name.

src/wolfgame.py
```diff
@@ -6972,11 +6980,13 @@ def player_stats(cli, nick, chan, rest):
     else:
         role = " ".join(params[1:])
         if role not in var.ROLE_GUIDE.keys():
-            match, _ = complete_match(role, var.ROLE_GUIDE.keys() | ["lover"])
+            match = complete_match(role, var.ROLE_GUIDE.keys() | ["lover"])
```
Vgr255: Need plural variable name here.

src/wolfgame.py
```diff
@@ -6972,11 +6980,13 @@ def player_stats(cli, nick, chan, rest):
...
             if not match:
                 reply(cli, nick, chan, messages["no_such_role"].format(role))
                 return
-            role = match
```
Vgr255: You're never grabbing the role again; need to do a `.pop()` at some point.

src/wolfgame.py
```diff
@@ -6972,11 +6980,13 @@ def player_stats(cli, nick, chan, rest):
...
+            if len(match) > 2:
```
Vgr255: What if there are exactly two matches?

src/utilities.py
```diff
@@ -411,18 +411,19 @@ def break_long_message(phrases, joinstr = " "):
...
-        return None, num_matches
-    else:
-        return bestmatch, 1
+            possible_matches.add(possible)
+    return sorted(possible_matches)
+
+def complete_one_match(string, matches):
+    match = complete_match(string,matches) 
+    if len(match) == 1:
+        return match.pop()
```
Vgr255: You have a list, so you can just index it - `return matches[0]` for instance. Also update all the other places that use `pop()`.

src/wolfgame.py
```diff
@@ -6994,13 +7004,16 @@ def vote_gamemode(var, wrapper, gamemode, doreply):
         return
 
     if gamemode not in var.GAME_MODES.keys():
-        match, _ = complete_match(gamemode, var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
-        if not match:
-            if doreply:
-                wrapper.pm(messages["invalid_mode_no_list"].format(gamemode))
+        matches = complete_match(gamemode, var.GAME_MODES.keys() - ["roles", "villagergame"] - var.DISABLED_GAMEMODES)
```
Vgr255: Style nit: change `["roles", "villagergame"]` for `{"roles", "villagergame"}` - i.e. curly braces instead of square brackets. Also update the lover one above.

src/wolfgame.py
```diff
@@ -7354,7 +7367,7 @@ def fgame(cli, nick, chan, rest):
 
             if gamemode not in var.GAME_MODES.keys() - var.DISABLED_GAMEMODES:
                 gamemode = gamemode.split()[0]
-                gamemode, _ = complete_match(gamemode, var.GAME_MODES.keys() - var.DISABLED_GAMEMODES)
+                gamemode = complete_match(gamemode, var.GAME_MODES.keys() - var.DISABLED_GAMEMODES)
```
Vgr255: You're getting a list as a response - didn't you want `complete_one_match` instead?

src/wolfgame.py
```diff
@@ -6994,13 +7014,16 @@ def vote_gamemode(var, wrapper, gamemode, doreply):
...
+        matches = complete_match(gamemode, var.GAME_MODES.keys() - {"roles", "villagergame"} - var.DISABLED_GAMEMODES)
+        if not matches:
+            wrapper.pm(messages["invalid_mode_no_list"].format(gamemode))
             return
-        gamemode = match
-
+        if len(matches) > 1:
+            wrapper.pm(messages["invalid_mode"].format(gamemode, ", ".join(matches)))
+            return
+        else:
```
nyuszika7h: The `else` is redundant here because of the `return`, just put the next line outside of the `if` block unindented. (In some cases `if`/`else` still makes for cleaner code even if return is used, but this is not one of those cases.)

nyuszika7h: The `else` is redundant here because of the `return`, just put the next line outside of the `if` block unindented. (In some cases `if`/`else` still makes for cleaner code even if return is used, but this is not one of those cases.)

src/wolfgame.py
```diff
@@ -7354,9 +7377,9 @@ def fgame(cli, nick, chan, rest):
 
             if gamemode not in var.GAME_MODES.keys() - var.DISABLED_GAMEMODES:
                 gamemode = gamemode.split()[0]
-                gamemode, _ = complete_match(gamemode, var.GAME_MODES.keys() - var.DISABLED_GAMEMODES)
+                gamemode = complete_one_match(gamemode, var.GAME_MODES.keys() - var.DISABLED_GAMEMODES)
                 if not gamemode:
-                    cli.notice(nick, messages["invalid_mode_no_list"].format(rest))
+                    cli.notice(nick, messages["invalid_mode_no_list"].format(rest[0]))
```
nyuszika7h: Why are you using `rest[0]` here? It doesn't seem like `rest` is modified, so that would return the first *character*. If you meant the first word, use `rest.split()[0]`.

nyuszika7h:
**Note to whoever merges:** Please squash and write a descriptive commit message instead of "Update functions".

jacob1:
merged locally in 0a6c6200abc34434571239b80b702744821db34b

PR #284, opened 2017-01-20
Split succubus

skizzerz:
It's been tested (not exhaustively, but enough to give me confidence everything works). There are a few behavior changes rolled into this compared to how succubus used to work; these are described in the commit message.


src/gamemodes.py
```diff
@@ -1221,35 +1216,10 @@ def on_join(self, evt, var, wrapper, message, forced=False):
 
     def _on_join(self, var, wrapper):
         role = random.choice(self.roles)
+        newlist = copy.deepcopy(var.ROLES)
+        newlist[role].add(wrapper.source.nick)
```
Vgr255: I'd like a FIXME comment here (since `var.ROLES` will hold User instances, eventually).

src/wolfgame.py
```diff
@@ -1800,33 +1799,23 @@ def chk_decision(cli, force=""):
...
                 # 4 = fool
                 # 5 = desperation totem, other things that happen on generic lynch
-                vote_evt = Event("chk_decision_lynch", {
-                    "votee": votee,
-                    "deadlist": deadlist
-                    }, del_player=del_player)
+                vote_evt = Event("chk_decision_lynch", {"votee": votee, "deadlist": deadlist},
+                    del_player=del_player, original_votee=votee,
+                    force=(votee == force), votelist=votelist,
+                    not_lynching=not_lynching)
```
nyuszika7h: You shouldn't wrap arguments like this according to PEP 8. It's harder to follow.

src/roles/succubus.py
```diff
@@ -0,0 +1,315 @@
...
+        s = nick if a == prefix else a
+        t = nick if b == prefix else b
+        kvp[s] = t
+    VISITED.update(kvp)
+    if prefix in VISITED:
+        del VISITED[prefix]
+
+@event_listener("reset")
+def on_reset(evt, var):
+    ALL_SUCC_IDLE = True
```
Vgr255: There needs to be a `global` statement here, otherwise this does nothing.

src/roles/succubus.py
```diff
@@ -0,0 +1,315 @@
...
+    evt.data["actedcount"] += len(VISITED)
+    evt.data["nightroles"].extend(var.ROLES["succubus"])
+
+@event_listener("targeted_command")
+def on_targeted_command(evt, cli, var, cmd, actor, orig_target, tags):
+    if "beneficial" not in tags and actor in ENTRANCED and evt.data["target"] in var.ROLES["succubus"]:
+        try:
+            what = evt.params.action
+        except AttributeError:
+            what = cmd
```
Vgr255: Please use a better variable name than `cmd` or `command`, to avoid confusion with the decorators.

src/roles/succubus.py
```diff
@@ -0,0 +1,315 @@
...
+        del VISITED[nick]
+
+    # if all succubi are dead, one of two things happen:
+    # 1. if all succubi idled out (every last one of them), un-entrance people
+    # 2. otherwise, kill all entranced people immediately, they still remain entranced (and therefore lose)
+    # death_triggers is False for an idle-out, so we use that to determine which it is
+    if death_triggers:
+        ALL_SUCC_IDLE = False
+    if len(var.ROLES["succubus"]) == 0:
+        if ALL_SUCC_IDLE:
```
Vgr255: This won't work - `ALL_SUCC_IDLE` has been assigned as a local variable, but if it hasn't been set to anything, it _will not_ fall back to the global one. Choices:

- Use a differently-named local variable, which you originally assign to the global version, then update that (If you don't want the global one to change)
- Add a `global` statement for the variable at the top of the function (If you wish to change the global variable to change)
- Use a mutable data structure, like a list, which you can then do stuff with to avoid the `global` shenanigans

src/roles/succubus.py
```diff
@@ -0,0 +1,315 @@
...
+            evt.data["message"].append(messages["target_not_home"])
+            evt.data["novictmsg"] = False
+        evt.stop_processing = True
+        evt.prevent_default = True
+
+@event_listener("transition_day_resolve_end", priority=1)
+def on_transition_day_resolve_end(evt, cli, var, victims):
+    for victim in victims + evt.data["bitten"]:
+        if victim in evt.data["dead"] and victim in VISITED.values() and (victim in evt.data["bywolves"] or victim in evt.data["bitten"]):
+            for succ in VISITED.keys():
```
Vgr255: No need to use `.keys()` here, just iterate over the dict. Or use `.items()` directly, thus meaning you don't need to access the dict again for the value.

src/roles/succubus.py
Vgr255: (Also see above comment about how to handle this variable)

src/roles/wolf.py
```diff
@@ -394,6 +394,33 @@ def on_transition_night_end(evt, cli, var):
...
+                var.FINAL_ROLES[wc] = "wolf"
+                pm(cli, wc, messages["cub_grow_up"])
+                debuglog(wc, "(wolf cub) GROW UP")
+    if did_something:
+        evt.prevent_default = True
+        evt.stop_processing = True
+
+@event_listener("succubus_visit")
+def on_succubus_visit(evt, cli, var, nick, victim):
+    if set(KILLS.get(victim, ())) & var.ROLES["succubus"]:
```
Vgr255: IMO, this is better: `if var.ROLES["succubus"].intersection(KILLS.get(victim, ())):`

src/utilities.py
```diff
@@ -309,15 +309,20 @@ def singular(plural):
     # otherwise we just added an s on the end
     return plural[:-1]
 
-def list_players(roles=None):
+def list_players(roles=None, rolemap=None):
```
Vgr255: If you're going to use this a keyword-only argument, please forcibly make it keyword-only, e.g. `def list_players(roles=None, *, rolemap=None):`

Vgr255:
Left several comments; some are mostly style nits, but others are critical and this shouldn't get merged until they're addressed.

src/roles/succubus.py
skizzerz: This is the only thing I'm not planning on changing, because the only other thing it would make sense to change it to is "command." The argument is the string command name that was called and passed to this event, for example "visit", "kill", or "target." cmd/command is therefore the most descriptive and accurate variable name, and as the class is not used in the function, shadowing that name is functionally fine.

Issue #228, opened 2016-06-22
!fjoin does not autocomplete nicks


nyuszika7h:

Fudster:
Do you want me show the possible matches?

Or just:
```
You may only fjoin people who are in this channel
```

Vgr255:
No need to show the possible nicks.

Fudster:
#285 Closes this.

PR #285, opened 2017-02-06
Tweak fjoin

Fudster:
This is for issue #228

Additionally discussed on IRC, Only allow joining multiple nicknames in debug mode.


src/wolfgame.py
```diff
@@ -979,14 +979,28 @@ def fjoin(var, wrapper, message):
         "join_deadchat": join_deadchat,
         "vote_gamemode": vote_gamemode
         })
+    ul = list(var.USERS.keys())
+    ull = [u.lower() for u in ul]
+
```
skizzerz: You moved this way too far up. Keep this closer to where you first use it, instead of interrupting the flow of a block of logic completely unrelated to these variables.

Vgr255:
Your PR highlighted some issues with `fjoin` that I should have already fixed before. I'll fix them in master, then you should rebase your work on top of it. There will be one minor conflict, but the bulk of your changes should remain unaffected.

skizzerz:
(Note that while my concerns were addressed, this PR will likely need some additional work after Vgr pushes his stuff)

src/wolfgame.py
```diff
@@ -979,16 +979,31 @@ def fjoin(var, wrapper, message):
...
         return
     noticed = False
     fake = False
     if not message.strip():
         evt.data["join_player"](var, wrapper, forced=True)
-
-    for tojoin in re.split(" +", message):
+    
+    ul = list(var.USERS.keys())
+    ull = [u.lower() for u in ul]
```
Vgr255: Remove this bit, it's no longer needed.

src/wolfgame.py
```diff
@@ -979,16 +979,31 @@ def fjoin(var, wrapper, message):
...
+    
+    if not botconfig.DEBUG_MODE:
+        message = re.split(" +", message)
+        match = complete_one_match(irc_lower(message[0]), ull)
```
Vgr255: Don't use `irc_lower`, use `users.lower` instead.

src/wolfgame.py
```diff
@@ -979,16 +979,31 @@ def fjoin(var, wrapper, message):
...
     if not message.strip():
         evt.data["join_player"](var, wrapper, forced=True)
-
-    for tojoin in re.split(" +", message):
+    
+    ul = list(var.USERS.keys())
+    ull = [u.lower() for u in ul]
+    
+    if not botconfig.DEBUG_MODE:
+        message = re.split(" +", message)
```
Vgr255: Use a different variable - `message` should remain the original string we got.

src/wolfgame.py
```diff
@@ -979,16 +979,31 @@ def fjoin(var, wrapper, message):
...
+        match = complete_one_match(irc_lower(message[0]), ull)
+        if match:
+            message = [match]
```
Vgr255: Same here - change variable name.

src/wolfgame.py
```diff
@@ -1002,6 +1017,7 @@ def fjoin(var, wrapper, message):
         if not tojoin:
             continue
 
+
```
Vgr255: Superfluous whitespace.

src/wolfgame.py
```diff
@@ -979,16 +979,29 @@ def fjoin(var, wrapper, message):
...
         evt.data["join_player"](var, wrapper, forced=True)
-
-    for tojoin in re.split(" +", message):
+   
+    parts = re.split(" +", message)
+    possible_users = {u.lower().nick for u in wrapper.target.users}
+    if not botconfig.DEBUG_MODE:
+        match = complete_one_match(users.lower(parts[0]), possible_users)
+        if match:
+            tojoin = [match]
```
Vgr255: Use a different variable name, even make it something like `to_join`, because I don't want the loop variable further down to be the same as the iterable.

src/wolfgame.py
```diff
@@ -2920,7 +2933,7 @@ def goat(cli, nick, chan, rest):
     if var.GOATED and nick not in var.SPECIAL_ROLES["goat herder"]:
         cli.notice(nick, messages["goat_fail"])
         return
-
+
```
Vgr255: Why is there random whitespace here?

PR #286, opened 2017-02-08
Goat revamp

### LINKED ISSUE #276: Revamp !goat
As I mentioned earlier on IRC, I believe the way `!goat` works needs to be revisited. Currently, it is pretty much completely useless except for the goat herder (which is the last person who votes to start), which results in `!fgoat` being handed out like candy. Even though we can remove the privileges of abusers, I would find it better to remove `!fgoat` entirely, and make `!goat` work in any phase (even outside games) but put a rate limit it, such as one goat every 5 minutes – either a global or a per-user rate limit, not sure which would be better yet. And possibly get rid of goat herder too.

And if people abuse it, we can simply deny it to them with a warning, like this:

```
!fwarn Example 1 deny=goat Abusing !goat
```

(This is just an example, the number of warning points depends on the situation. We may give out a 0-point warning instead if the user has many warning points and we determine stasis is not deserved.)

Fudster:
Closes #276

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -2945,15 +2945,14 @@ def goat(var, wrapper, message):
         wrapper.pm(messages["goat_target_not_in_channel"].format(target))
         return
 
-    var.LAST_GOAT[wrapper.source.nick] = [datetime.now(), 1]
+    var.LAST_GOAT[wrapper.source] = [datetime.now(), 1]
     goatact = random.choice(messages["goat_actions"])
-    wrapper.send(messages["goat_success"].format(
-        wrapper.source, goatact, victim))
+    wrapper.send(messages["goat_success"].format(wrapper.source, goatact, victim))
 
 @command("fgoat", flag="j")
 def fgoat(var, wrapper, message):
     """Forces a goat to interact with anyone or anything, without limitations."""
-    
+
     nick = message.split(' ')[0].strip()
     possible_users = {u.lower().nick for u in wrapper.target.users}
     victim = complete_one_match(users.lower(nick), possible_users)
@@ -3346,7 +3345,7 @@ def begin_day(cli):
     var.DISEASED = set()
     var.MISDIRECTED = set()
     var.DYING = set()
-    var.LAST_GOAT = {}
+    var.LAST_GOAT.clear()
     msg = messages["villagers_lynch"].format(botconfig.CMD_CHAR, len(list_players()) // 2 + 1)
     cli.msg(chan, msg)
 
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -3409,7 +3409,6 @@ def transition_day(cli, gameid=0):
         return
 
     var.PHASE = "day"
-    var.GOATED = False
     var.DAY_COUNT += 1
     var.FIRST_DAY = (var.DAY_COUNT == 1)
     var.DAY_START_TIME = datetime.now()
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -2928,33 +2928,24 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
 #    if "@" + botconfig.NICK in names:
 #        var.OPPED = True
 
-@cmd("goat", playing=True, phases=("day",))
-def goat(cli, nick, chan, rest):
+@command("goat")
+def goat(var, wrapper, message):
     """Use a goat to interact with anyone in the channel during the day."""
-
-    if var.GOATED and nick not in var.SPECIAL_ROLES["goat herder"]:
-        cli.notice(nick, messages["goat_fail"])
-        return
-
-    ul = list(var.USERS.keys())
-    ull = [x.lower() for x in ul]
-
-    rest = re.split(" +",rest)[0]
-    if not rest:
-        cli.notice(nick, messages["not_enough_parameters"])
-
-    victim = complete_one_match(rest.lower(), ull)
+    
+    victim = re.split(" +",message)[0]
     if not victim:
-        cli.notice(nick, messages["goat_target_not_in_channel"].format(rest))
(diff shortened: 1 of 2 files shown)


src/wolfgame.py
```diff
@@ -65,6 +65,7 @@
 var.LAST_TIME = None
 var.LAST_START = {}
 var.LAST_WAIT = {}
+var.LAST_GOAT = {}
 
 var.USERS = {}
```
Vgr255: Just below this line, there is a `var.SPECIAL_ROLES` dict being created - you can remove that line.

src/wolfgame.py
```diff
@@ -2928,45 +2929,43 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
 
-@cmd("goat", playing=True, phases=("day",))
-def goat(cli, nick, chan, rest):
+@command("goat")
+def goat(var, wrapper, message):
     """Use a goat to interact with anyone in the channel during the day."""
-
-    if var.GOATED and nick not in var.SPECIAL_ROLES["goat herder"]:
-        cli.notice(nick, messages["goat_fail"])
+
```
Vgr255: Extraneous whitespace.

src/wolfgame.py
```diff
@@ -2928,45 +2929,43 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
-@cmd("goat", playing=True, phases=("day",))
-def goat(cli, nick, chan, rest):
+@command("goat")
+def goat(var, wrapper, message):
     """Use a goat to interact with anyone in the channel during the day."""
-
-    if var.GOATED and nick not in var.SPECIAL_ROLES["goat herder"]:
-        cli.notice(nick, messages["goat_fail"])
+    
+    if wrapper.source.nick in var.LAST_GOAT and var.LAST_GOAT[wrapper.source.nick][0] + timedelta(seconds=var.GOAT_RATE_LIMIT) > datetime.now():
```
Vgr255: Make `var.LAST_GOAT` hold `User` instances instead of strings.

src/wolfgame.py
```diff
@@ -2928,45 +2929,43 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
-    rest = re.split(" +",rest)[0]
-    if not rest:
-        cli.notice(nick, messages["not_enough_parameters"])
-
-    victim = complete_one_match(rest.lower(), ull)
+    target = re.split(" +",message)[0]
+    if not target:
+        wrapper.pm(messages["not_enough_parameters"])
+    
+
```
Vgr255: Both of those lines have extra whitespace, and you're missing a return.

src/wolfgame.py
```diff
@@ -2928,45 +2929,43 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
+    
+    possible_users = {u.lower().nick for u in wrapper.target.users}
+    victim = complete_one_match(users.lower(target), possible_users)
     if not victim:
-        cli.notice(nick, messages["goat_target_not_in_channel"].format(rest))
+        wrapper.pm(messages["goat_target_not_in_channel"].format(target))
         return
-    victim = ul[ull.index(victim)]
-
+
```
Vgr255: Whitespace!

src/wolfgame.py
```diff
@@ -2928,45 +2929,43 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
-
-@cmd("fgoat", flag="j")
-def fgoat(cli, nick, chan, rest):
+@command("fgoat", flag="j")
+def fgoat(var, wrapper, message):
     """Forces a goat to interact with anyone or anything, without limitations."""
-    nick_ = rest.split(' ')[0].strip()
-    if nick_.lower() in (x.lower() for x in users.users()):
-        togoat = nick_
+
```
Vgr255: Some more whitespace.

src/wolfgame.py
skizzerz: Also remove every instance that dict is mentioned/used. You missed some spots elsewhere, like in game start.

Vgr255: It's actually removed in game start.

src/wolfgame.py
```diff
@@ -2928,45 +2928,42 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
+        return
+    possible_users = {u.lower().nick for u in wrapper.target.users}
+    victim = complete_one_match(users.lower(target), possible_users)
     if not victim:
-        cli.notice(nick, messages["goat_target_not_in_channel"].format(rest))
+        wrapper.pm(messages["goat_target_not_in_channel"].format(target))
         return
-    victim = ul[ull.index(victim)]
 
+    var.LAST_GOAT[wrapper.source.nick] = [datetime.now(), 1]
```
Vgr255: You're storing the nick here, instead of the instance.

src/wolfgame.py
```diff
@@ -2928,45 +2928,42 @@ def on_join(cli, raw_nick, chan, acc="*", rname=""):
...
     goatact = random.choice(messages["goat_actions"])
+    wrapper.send(messages["goat_success"].format(
+        wrapper.source, goatact, victim))
```
Vgr255: Needless wrapping of arguments; put them on the same line for clarity.

src/wolfgame.py
```diff
@@ -3349,7 +3346,7 @@ def begin_day(cli):
     var.DISEASED = set()
     var.MISDIRECTED = set()
     var.DYING = set()
-
+    var.LAST_GOAT = {}
```
Vgr255: What is the purpose of setting this variable here back to an empty dict? Also, since the dict already exists, this should be `var.LAST_GOAT.clear()`. That way we keep the same dictionnary for the entire span of the process.

PR #291, opened 2017-03-24
Add a new swap_player event

Vgr255:
What it says on the title. Fairly simple, does not touch much. Making everything else use it is not the goal of this PR. I just want to get this in and working so that future stuff can use it easily.


src/wolfgame.py
```diff
@@ -568,14 +568,20 @@ def replace(var, wrapper, message):
         target, _ = users.complete_match(rest[0], pl)
 
         if target is None:
-            wrapper.pm(messages["target_not_playing"].format(" longer" if target in var.DEAD else "t"))
+            wrapper.pm(messages["target_not_playing"].format("t"))
```
skizzerz: Split this message up into two keys, one with "not playing" and one with "no longer playing"

The way it is right now is fugly and needs to die

Vgr255: Agreed.

Vgr255: Done

PR #292, opened 2017-03-25
Convert dullahan to use the new API

Vgr255:
Tested thanks to @nyuszika7h, working as intended. This is the first iteration of converting roles to use the new API, more will follow!


src/roles/dullahan.py
```diff
@@ -5,76 +5,73 @@
 
 import src.settings as var
```
skizzerz: Remove this import as it is no longer needed

src/roles/dullahan.py
```diff
@@ -5,76 +5,73 @@
...
         return
-    victim = evt.data["target"]
+    target = users._get(evt.data["target"]) # FIXME: Need to fix once targeted_command uses the new API
 
-    KILLS[nick] = victim
+    KILLS[wrapper.source] = target
 
-    msg = messages["wolf_target"].format(orig)
-    pm(cli, nick, messages["player"].format(msg))
+    wrapper.pm(messages["player"].format(messages["wolf_target"].format(orig)))
```
skizzerz: This is also gross. Don't stitch together multiple messages. Make a new message for "You have selected {0} to be killed." vs "{0} has selected {1} to be killed." Both should be generic (i.e. not reference "wolf" in the key, as not only wolf kills now)

You don't need to update wolf.py as part of this PR to use the more sane messaging, but do update the above instance. You can if you want though (hunter and vigilante probably need it too).

src/roles/dullahan.py
```diff
@@ -5,76 +5,73 @@
...
 
-    debuglog("{0} ({1}) KILL: {2} ({3})".format(nick, get_role(nick), victim, get_role(victim)))
+    debuglog("{0} ({1}) KILL: {2} ({3})".format(wrapper.source, get_role(wrapper.source.nick), target, get_role(target.nick)))
```
skizzerz: change get_role(wrapper.source.nick) to a hardcoded string "dullahan"

If dullahan is applied as a template this would cause very wacky debug logs. This will be something that gets adjusted going forwards; debug logs should always hardcode the roles that are triggering the logs, as such roles are already known.

src/roles/dullahan.py
```diff
@@ -5,76 +5,73 @@
...
+        if v.nick == nick:
+            h.send(messages["hunter_discard"])
             del KILLS[h]
-        elif h == nick:
+        elif h.nick == nick:
             del KILLS[h]
     if death_triggers and nickrole == "dullahan":
         pl = evt.data["pl"]
-        targets = TARGETS[nick] & set(pl)
+        targets = TARGETS[users._get(nick)].union(users._get(x) for x in pl) # FIXME
```
skizzerz: Why did you change this to a union? The original code is an intersection. Union is very wrong here.

src/roles/dullahan.py
```diff
@@ -103,100 +100,88 @@ def on_del_player(evt, cli, var, nick, nickrole, nicktpls, death_triggers):
...
-        del KILLS[actor]
-    if nick in KILLS:
-        del KILLS[nick]
-    if actor_role == "dullahan" and nick_role != "dullahan" and actor in TARGETS:
-        TARGETS[nick] = TARGETS[actor] - {nick}
-        del TARGETS[actor]
-    elif nick_role == "dullahan" and actor_role != "dullahan" and nick in TARGETS:
-        TARGETS[actor] = TARGETS[nick] - {actor}
-        del TARGETS[nick]
+    for k in set(KILLS):
```
skizzerz: This is gross. if you don't want to check-then-del, then use KILLS.pop(actor, None) and KILLS.pop(nick, None). Don't turn an O(1) operation into O(n).

src/roles/dullahan.py
```diff
@@ -103,100 +100,88 @@ def on_del_player(evt, cli, var, nick, nickrole, nicktpls, death_triggers):
...
+        if k.nick == actor or k.nick == nick:
+            del KILLS[k]
+
+    for k in set(TARGETS):
```
skizzerz: Same here, don't make this O(n) when it doesn't need to be.

src/roles/dullahan.py
```diff
@@ -103,100 +100,88 @@ def on_del_player(evt, cli, var, nick, nickrole, nicktpls, death_triggers):
...
 @event_listener("transition_night_end", priority=2)
 def on_transition_night_end(evt, cli, var):
-    for dullahan in var.ROLES["dullahan"]:
-        targets = list(TARGETS[dullahan])
-        for target in var.DEAD:
-            if target in targets:
-                targets.remove(target)
-        if not targets: # already all dead
-            pm(cli, dullahan, "{0} {1}".format(messages["dullahan_simple"], messages["dullahan_targets_dead"]))
+    for dullahan, targets in TARGETS.items():
```
skizzerz: Keep this iterating over var.ROLES["dullahan"] -- we want this to break if someone manages to be a dullahan without having an entry in TARGETS. Not doing this would result in silent and harder-to-track-down bugs (chk_nightdone is fine as-is though, don't need to change that back).

src/roles/dullahan.py
```diff
@@ -103,100 +100,88 @@ def on_del_player(evt, cli, var, nick, nickrole, nicktpls, death_triggers):
...
+        if not targs: # already all dead
+            dullahan.send("{0} {1}".format(messages["dullahan_simple"], messages["dullahan_targets_dead"]))
             continue
-        random.shuffle(targets)
-        if dullahan in var.PLAYERS and not is_user_simple(dullahan):
-            pm(cli, dullahan, messages["dullahan_notify"])
-        else:
-            pm(cli, dullahan, messages["dullahan_simple"])
+        random.shuffle(targs)
+        dullahan.send(messages["dullahan_" + ("simple" if dullahan.prefers_simple() else "notify")])
```
skizzerz: It was more readable before. Saving 3 lines is not an excuse to introduce spurious ternary conditionals. Plus, this way makes it impossible to search for the strings "dullahan_simple" and "dullahan_notify" to see where those message keys are being used.

Convert back to full if/else block, each calling dullahan.send() with the proper key.

src/roles/dullahan.py
```diff
@@ -103,100 +100,88 @@ def on_del_player(evt, cli, var, nick, nickrole, nicktpls, death_triggers):
...
         t = messages["dullahan_targets"] if var.FIRST_NIGHT else messages["dullahan_remaining_targets"]
-        pm(cli, dullahan, t + ", ".join(targets))
+        dullahan.send(t + ", ".join([t.nick for t in targs]))
```
skizzerz: I think str.join() just needs an iterable, so you don't need to use a list comprehension here, use a generator expression instead.

src/roles/dullahan.py
```diff
@@ -205,37 +190,47 @@ def on_role_assignment(evt, cli, var, gamemode, pl, restart):
...
+    if user in TARGETS:
+        succ_target = False
+        for target in set(TARGETS[user]):
+            if target.nick in var.ROLES["succubus"]:
+                TARGETS[user].remove(target)
+                succ_target = True
+        if succ_target:
+            pm(cli, victim, messages["dullahan_no_kill_succubus"])
+    if user in KILLS and KILLS[user].nick in var.ROLES["succubus"]:
+        pm(cli, victim, messages["no_kill_succubus"].format(KILLS[user]))
```
skizzerz: change format() to KILLS[user].nick

src/roles/dullahan.py
```diff
@@ -205,37 +190,47 @@ def on_role_assignment(evt, cli, var, gamemode, pl, restart):
...
+        del KILLS[user]
 
 @event_listener("myrole")
 def on_myrole(evt, cli, var, nick):
     role = get_role(nick)
```
skizzerz: check if nick is in var.ROLES["dullahan"] instead of checking if get_role(nick) returns "dullahan"

src/roles/dullahan.py
```diff
@@ -205,37 +190,47 @@ def on_role_assignment(evt, cli, var, gamemode, pl, restart):
...
-            if target in targets:
+        targets = list(TARGETS[users._get(nick)]) # FIXME
+        for target in list(targets):
+            if target.nick in var.DEAD:
                 targets.remove(target)
         random.shuffle(targets)
         if targets:
             t = messages["dullahan_targets"] if var.FIRST_NIGHT else messages["dullahan_remaining_targets"]
-            evt.data["messages"].append(t + ", ".join(targets))
+            evt.data["messages"].append(t + ", ".join([t.nick for t in targets]))
```
skizzerz: Again, join doesn't need a list comprehension users, genexpr preferred

src/roles/dullahan.py
```diff
@@ -205,37 +190,47 @@ def on_role_assignment(evt, cli, var, gamemode, pl, restart):
...
-        targets = TARGETS[nickname] - var.DEAD
+    user = users._get(nickname) # FIXME
+    if role == "dullahan" and user in TARGETS: # FIXME
+        targets = set(TARGETS[user])
+        for target in TARGETS[user]:
+            if target.nick in var.DEAD:
+                targets.remove(target)
         if targets: 
-            evt.data["special_case"].append("need to kill {0}".format(", ".join(TARGETS[nickname] - var.DEAD)))
+            evt.data["special_case"].append("need to kill {0}".format(", ".join([t.nick for t in targets])))
```
skizzerz: Again, use genexpr. Split hardcoded string into new message key (here as well as below for "all targets dead").

src/roles/dullahan.py
Vgr255: Except that's not actually possible - `KILLS` doesn't store nicks, it stores users. Therefore, we need to check for that. When that event is updated to use the new API, we won't need to do that anymore, but for now, we do. And nope, shoe-horning in a few `users._get` calls won't work - that function is O(n), and you'd probably call it several times.

src/roles/dullahan.py
Vgr255: See above comment.

src/roles/dullahan.py
Vgr255: Good question - probably just an oversight. Fixed.

src/roles/dullahan.py
Vgr255: Sure. It's going to be a bit hacky, but that can be done easily.

src/roles/dullahan.py
Vgr255: CPython implementation detail: If the iterable is not a list, it's implicitly converted to a list before performing the join. Yeah, it's an implementation detail, but I think that ultimately it doesn't matter.

src/roles/dullahan.py
Vgr255: Why? User instances work just as fine in `format` calls.

src/roles/dullahan.py
Vgr255: Fair enough.

src/roles/dullahan.py
skizzerz: My bad, I was assuming format called __str__ which would return the full hostmask (mostly because __format__ is defined in the base class, so it wasn't immediately obvious)

src/roles/dullahan.py
skizzerz: We don't rely on implementation details, see also: the whole weakref debacle.

PR #295, opened 2017-08-15
Split Mad Scientist

skizzerz:
Related changes:
- MS now honors all protections instead of only caring about blessed
  villagers
- Allow FA to bypass protection even if FA is applied as a template
- Make bodyguard swap themselves in as the assassination target rather
  than blocking the attempt outright and suiciding. This means that
  active protections on the bodyguard have a chance to save them as well.
- Redo some messages to make the above sound nice in the bot.
- Add some additional params to the assassinate event to see WHY the
  assassination is happening (source) and WHO is doing the assassination
  (killer; currently a nick but should be a user sometime in the future).
- Add a target data item to teh assassinate event so that listeners can
  change who is being assassinated. Protection boilerplate has been
  adjusted to account for this.


messages/en.json
```diff
@@ -209,6 +209,9 @@
     "mad_scientist_kill_single": "\u0002{0}\u0002 throws a potent chemical concoction into the crowd. \u0002{1}\u0002, a{2} \u0002{3}\u0002, gets hit by the chemicals and dies.",
     "mad_scientist_kill_single_no_reveal": "\u0002{0}\u0002 throws a potent chemical concoction into the crowd. \u0002{1}\u0002 gets hit by the chemicals and dies.",
     "mad_scientist_fail": "\u0002{0}\u0002 throws a potent chemical concoction into the crowd. Thankfully, nobody seems to have gotten hit.",
+    "mad_scientist_fail_totem": "Sensing danger, \u0002{1}\u0002's totem emits a brilliant flash of light, teleporting them away from \u0002{0}\u0002.",
+    "mad_scientist_fail_angel": "Sensing danger, a guardian angel whisks \u0002{1}\u0002 away from \u0002{0}\u0002.",
+    "mad_scientist_fail_bodyguard": "Sensing danger, \u0002{2}\u0002 shoves \u0002{1}\u0002 aside to save them from \u0002{0}\u0002.",
```
Vgr255: Same line under two different entries; should we keep it?

src/roles/madscientist.py
```diff
@@ -0,0 +1,176 @@
...
+    # if those players are already dead, nothing happens
+    for index, user in enumerate(var.ALL_PLAYERS):
+        if user.nick == nick: # FIXME
+            break
+
+    num_players = len(var.ALL_PLAYERS)
+    pl = evt.data["pl"]
+    targets = []
+    target1 = var.ALL_PLAYERS[index - 1]
+    target2 = var.ALL_PLAYERS[(index + 1) % num_players]
```
Vgr255: Huh, clever.

PR #296, opened 2017-08-30
Convert the del_player event

Vgr255:


src/roles/dullahan.py
```diff
@@ -59,16 +59,16 @@ def on_player_win(evt, var, user, role, winner, survived):
...
     for h, v in list(KILLS.items()):
-        if v.nick == nick:
+        if v is user:
             h.send(messages["hunter_discard"])
             del KILLS[h]
-        elif h.nick == nick:
+        elif h is user:
             del KILLS[h]
     if death_triggers and "dullahan" in allroles:
         pl = evt.data["pl"]
```
skizzerz: evt.data["pl"] should be updated to be a user list as well (this will require modifications in wolfgame.py as well in del_player and refresh_pl)

src/roles/dullahan.py
```diff
@@ -79,15 +79,15 @@ def on_del_player(evt, cli, var, nick, mainrole, allroles, death_triggers):
                     refresh_pl=evt.params.refresh_pl,
                     message_prefix="dullahan_die_",
                     source="dullahan",
-                    killer=nick,
+                    killer=user.nick,
```
skizzerz: update killer to be the user, not the nick

PR #297, opened 2017-09-04
Convert transition_day

Vgr255:
Convert most of the transition_day function, and all of its events (as well as all the events it eventually fires) to no longer use cli and be user-aware. Has been fully tested thanks to Iciloo. Awaiting @skizzerz's review.

### DIFF
--- a/src/roles/vengefulghost.py
+++ b/src/roles/vengefulghost.py
@@ -5,12 +5,12 @@ from collections import defaultdict
 import src.settings as var
 from src.utilities import *
 from src import channels, users, debuglog, errlog, plog
-from src.functions import get_players
+from src.functions import get_players, get_target, get_main_role
 from src.decorators import command, event_listener
 from src.messages import messages
 from src.events import Event
 
-KILLS = {} # type: Dict[str, str]
+KILLS = {} # type: Dict[users.User, users.User]
 GHOSTS = {} # type: Dict[users.User, str]
 
 # temporary holding variable, only non-empty during transition_day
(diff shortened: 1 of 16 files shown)


src/roles/vengefulghost.py
```diff
@@ -105,28 +105,15 @@ def on_del_player(evt, var, user, nickrole, nicktpls, death_triggers):
         user.send(messages["vengeful_turn"].format(GHOSTS[user]))
         debuglog(user.nick, "(vengeful ghost) TRIGGER", GHOSTS[user])
 
-@event_listener("rename_player")
```
skizzerz: Missing swap_player listener (not sure it does anything with VG right now since they're dead and we still have all of those dead player user tracking bugs, but couldn't hurt to add it in for when said bugs get fixed)

src/roles/vengefulghost.py
```diff
@@ -105,28 +105,15 @@ def on_del_player(evt, var, user, nickrole, nicktpls, death_triggers):
...
+    wolves = set(get_players(var.WOLFTEAM_ROLES))
+    villagers = set(get_players()) - wolves
     for ghost, target in GHOSTS.items():
         if target[0] == "!" or ghost.nick in var.SILENCED:
             continue
-        if ghost.nick not in KILLS:
+        if ghost not in KILLS:
             choice = set()
             if target == "wolves":
                 choice = wolves.copy()
```
skizzerz: Why are we doing .copy() here? (and 2 lines below). I know this wasn't your thing, but the copy seems superfluous.

src/roles/vengefulghost.py
```diff
@@ -136,39 +123,38 @@ def on_transition_day_begin(evt, cli, var):
...
         # important, otherwise our del_player listener messages the vg
         del KILLS[k]
 
-@event_listener("retribution_kill", priority=6) # FIXME: This function, and all of the event
-def on_retribution_kill(evt, cli, var, victim, orig_target):
-    t = evt.data["target"]
-    user = users._get(t)
+@event_listener("retribution_kill", priority=6)
+def on_retribution_kill(evt, var, victim, orig_target):
+    user = evt.data["target"]
```
skizzerz: var name, use "target"

src/roles/vigilante.py
```diff
@@ -97,16 +97,18 @@ def on_get_special(evt, var):
...
 
 @event_listener("transition_day", priority=2)
-def on_transition_day(evt, cli, var):
+def on_transition_day(evt, var):
     for k, d in list(KILLS.items()):
-        evt.data["victims"].append(d)
-        evt.data["onlybywolves"].discard(d)
-        evt.data["killers"][d].append(k)
+        actor = users._get(k) # FIXME
+        user = users._get(d) # FIXME
```
skizzerz: var names: killer, victim or killer, target

src/wolfgame.py
```diff
@@ -3646,35 +3648,36 @@ def transition_day(cli, gameid=0):
...
+                var.ROLES["lycan"].discard(victim.nick) # in the event lycan was a template, we want to ensure it gets purged
+                wolves = get_players(var.WOLFCHAT_ROLES)
                 random.shuffle(wolves)
                 wolves.remove(victim)  # remove self from list
-                for i, wolf in enumerate(wolves):
-                    pm(cli, wolf, messages["lycan_wc_notification"].format(victim))
-                    role = get_role(wolf)
+                to_send = []
+                for wolf in wolves:
+                    wolf.send(messages["lycan_wc_notification"].format(victim))
```
skizzerz: this message should be queued/bulk-sent (relay_wolfchat_command could result in it being sent twice, so don't use that wrapper here)

PR #301, opened 2017-10-22
Support for TLS certificate verification and client certificates.

miniCruzer:
If a TLS certificate fingerprint is provided, the client will check it against the SHA256 hex digest of the server's certificate. Different hash algorithms can be specified, and multiple fingerprints can be specified for networks with more than one server.

Not sure if `print()` is the right thing to do within the `connect()` method in order to display to the user when certificate validation fails.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 __pycache__/
 *.py[co]
 
+# linter files
+.mypy_cache/
+
 # Config files
 botconfig.py
 /gamemodes.py
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@
 __pycache__/
 *.py[co]
 
-# linter files
+# Linter files
 .mypy_cache/
 
 # Config files
--- a/oyoyo/client.py
+++ b/oyoyo/client.py
@@ -185,11 +185,18 @@ class IRCClient:
                 if self.cert_verify or self.cert_fp:
                     ctx.verify_mode = ssl.CERT_REQUIRED
                     ctx.load_default_certs()
+                else:
+                    self.stream_handler("NOT validating the servers TLS certificate! Set 'SSL_VERIFY=True' in botconfig.py to enable this.", level="warning")
 
                 if self.client_certfile:
                     # if client_keyfile is not specified, the ssl module will look to the
                     # client_certfile for it.
-                    ctx.load_cert_chain(self.client_certfile, self.client_keyfile)
+
+                    try:
+                        ctx.load_cert_chain(self.client_certfile, self.client_keyfile)
+                        self.stream_handler("Connecting with a TLS client certificate", level="info")
+                    except Exception as error:
+                        self.stream_handler("Unable to load client cert/key pair: {0}".format(error), level="warning")
 
                 try:
                     self.socket = ctx.wrap_socket(self.socket)
@@ -198,7 +205,6 @@ class IRCClient:
                     raise
 
                 if self.cert_fp:
-
                     algo = "sha256"
                     if self.cert_fp.find(":") != -1:
                         algo, fp = self.cert_fp.split(":")
@@ -206,9 +212,11 @@ class IRCClient:
                     else:
                         fp = self.cert_fp.split(",")
 
+                    self.stream_handler("Checking server's certificate {0} hash sum".format(algo), level="info")
+
                     try:
                         h = hashlib.new(algo)
-                    except ValueError as error:
+                    except Exception as error:
                         self.stream_handler("TLS certificate fingerprint verification failed: {}".format(error), level="warning")
                         self.stream_handler("Supported algorithms on this sytem: {0}".format(", ".join(hashlib.algorithms_available)), level="warning")
                         raise
@@ -224,6 +232,7 @@ class IRCClient:
                     if not matched:
                         self.stream_handler("Certificate fingerprint {0} did not match any excpected fingerprints".format(peercertfp), level="warning")
                         raise ssl.CertificateError("Certificate fingerprint does not match.")
+                    self.stream_handler("Server certificate fingerprint matched {0}".format(matched), level="info")
 
                 if self.cert_verify and not self.cert_fp:
                     cert = self.socket.getpeercert()
@@ -234,6 +243,7 @@ class IRCClient:
                         self.stream_handler("TLS certificate verification failed: {0}".format(error), level="warning")
                         raise
 
+                    self.stream_handler("Server's certificate is trusted.", level="info")
 
             if not self.blocking:
                 self.socket.setblocking(0)
--- a/oyoyo/client.py
+++ b/oyoyo/client.py
@@ -203,7 +203,7 @@ class IRCClient:
                     # TLS session tickets harm forward secrecy (this symbol is only defined in 3.6 and later)
                     ctx.options |= ssl.OP_NO_TICKET
 
-                if self.cert_verify or self.cert_fp:
+                if self.cert_verify and not self.cert_fp:
                     ctx.verify_mode = ssl.CERT_REQUIRED
 
                     if not self.cert_fp:
@@ -211,8 +211,8 @@ class IRCClient:
 
                     ctx.load_default_certs()
 
-                else:
-                    self.stream_handler("NOT validating the server's TLS certificate! Set 'SSL_VERIFY=True' in botconfig.py to enable this.", level="warning")
+                elif not self.cert_verify and not self.cert_fp:
+                    self.stream_handler("NOT validating the server's TLS certificate! Set 'SSL_VERIFY=True' or define a fingerprint in 'SSL_CERTFP' in botconfig.py to enable this.", level="warning")
 
                 if self.client_certfile:
                     # if client_keyfile is not specified, the ssl module will look to the
--- a/oyoyo/client.py
+++ b/oyoyo/client.py
@@ -193,8 +193,8 @@ class IRCClient:
 
                 try:
                     self.socket = ctx.wrap_socket(self.socket)
-                except (ssl.CertificateError, ssl.SSLError, Exception) as error:
-                    print("Error occured while connecting with TLS: {0}".format(error))
+                except Exception as error:
+                    self.stream_handler("Error occured while connecting with TLS: {0}".format(error), level="warning")
                     raise
 
                 if self.cert_fp:
@@ -209,8 +209,8 @@ class IRCClient:
                     try:
                         h = hashlib.new(algo)
                     except ValueError as error:
-                        print("TLS certificate fingerprint verification failed: {}".format(error))
-                        print("Supported algorithms on this sytem: {0}".format(", ".join(hashlib.algorithms_available)))
+                        self.stream_handler("TLS certificate fingerprint verification failed: {}".format(error), level="warning")
+                        self.stream_handler("Supported algorithms on this sytem: {0}".format(", ".join(hashlib.algorithms_available)), level="warning")
                         raise
 
                     h.update(self.socket.getpeercert(True))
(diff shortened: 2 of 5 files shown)


oyoyo/client.py
```diff
@@ -174,7 +180,60 @@ def connect(self):
...
+                    ctx.load_default_certs()
+
+                if self.client_certfile:
+                    # if client_keyfile is not specified, the ssl module will look to the
+                    # client_certfile for it.
+                    ctx.load_cert_chain(self.client_certfile, self.client_keyfile)
+
+                try:
+                    self.socket = ctx.wrap_socket(self.socket)
+                except (ssl.CertificateError, ssl.SSLError, Exception) as error:
```
Vgr255: `ssl.CertificateError` and `ssl.SSLError` both inherit from `Exception` -- either you catch only `Exception` (as in, any error whatsoever) or you don't :)

oyoyo/client.py
```diff
@@ -174,7 +180,60 @@ def connect(self):
...
+                    print("Error occured while connecting with TLS: {0}".format(error))
```
Vgr255: Please use `self.stream_handler` instead of `print` as it has more useful logging.

miniCruzer:
Requested changes made, plus some additional error handling. 

Didn't mean to push 46f2cd8, was bored on a 22 hour flight and forgot it was in there (reverted by 3b61e48).

miniCruzer:
Altogether, this accomplishes:

- Refuses to connect to a server with self signed certificates with`SSL_VERIFY = True` and `SSL_CERTFP` undefined.
- Refuses to connect to a server with a trusted certificate if the hostname listed in botconfig.py does not match the name on the certificate with `SSL_VERIFY = True` and `SSL_CERTFP` undefined.
- If `SSL_CERTFP` is defined, DOES NOT attempt to validate certificate trust (regardless of `SSL_VERIFY`'s value); instead, refuses to connect if the hash of the server certificate does not match one of the listed fingerprints.
- If `SSL_CERTFILE` is defined, will connect with a client certificate in this file. If `SSL_KEYFILE` is undefined, the client will attempt to load the key from `SSL_CERTFILE` instead. Upon failure, the client will connect anyway.
- If `SSL_CIPHERS` is defined, the client will only support negotiating the listed ciphers. Admins can view supported ciphers on their system with `$ openssl ciphers`. If an invalid cipher is selected or the server cannot negotiate a common cipher with the client, the connection will fail.
- Explicitly disables the broken/outdated protocols SSLv2, SSLv3, and TLSv1. Disables compression, and if running on Python 3.6 or later, disables TLS session tickets.

oyoyo/client.py
```diff
@@ -174,7 +181,86 @@ def connect(self):
...
+
+                try:
+                    self.socket = ctx.wrap_socket(self.socket, server_hostname=self.host)
+                except Exception as error:
+                    self.stream_handler("Error occured while connecting with TLS: {0}".format(error), level="warning")
+                    raise
+
+                if self.cert_fp:
+                    algo = "sha256"
+                    if self.cert_fp.find(":") != -1:
```
nyuszika7h: You can just do `if ":" in self.cert_fp:`.

oyoyo/client.py
```diff
@@ -174,7 +181,86 @@ def connect(self):
...
+                    else:
+                        fp = self.cert_fp.split(",")
+
+                    self.stream_handler("Checking server's certificate {0} hash sum".format(algo), level="info")
+
+                    try:
+                        h = hashlib.new(algo)
+                    except Exception as error:
+                        self.stream_handler("TLS certificate fingerprint verification failed: {}".format(error), level="warning")
+                        self.stream_handler("Supported algorithms on this sytem: {0}".format(", ".join(hashlib.algorithms_available)), level="warning")
```
nyuszika7h: Typo: sytem -> system

oyoyo/client.py
```diff
@@ -174,7 +181,86 @@ def connect(self):
...
+                    h.update(self.socket.getpeercert(True))
+                    peercertfp = h.hexdigest()
+
+                    matched = False
+                    for fingerprint in fp:
+                        if hmac.compare_digest(fingerprint, peercertfp):
+                            matched = fingerprint
+
+                    if not matched:
+                        self.stream_handler("Certificate fingerprint {0} did not match any excpected fingerprints".format(peercertfp), level="warning")
```
nyuszika7h: Typo: excpected -> expected

oyoyo/client.py
```diff
@@ -174,7 +181,86 @@ def connect(self):
...
+                        self.stream_handler("Unable to load client cert/key pair: {0}".format(error), level="warning")
+
+                try:
+                    self.socket = ctx.wrap_socket(self.socket, server_hostname=self.host)
+                except Exception as error:
+                    self.stream_handler("Error occured while connecting with TLS: {0}".format(error), level="warning")
+                    raise
+
+                if self.cert_fp:
+                    algo = "sha256"
```
nyuszika7h: Perhaps this should be auto-detected. Shouldn't be too hard to do based on length once you strip colons.

src/settings.py
```diff
@@ -6,6 +6,20 @@
...
+# you may specify the hash algorithm to use when verifying fingerprints. if unspecified, it will
+# default to SHA256. multiple fingerprints may be comma separated for networks with multiple
+# servers. check your Python version for suported algorithms.
+# syntax: SSL_CERTFP = "[HASH-ALGO]:fingerprint1,fingerprint2,fingerprint3"
+SSL_VERIFY = True
+SSL_CERTFILE = None # client cert file to connect with in PEM format; can also contain keyfile
+SSL_KEYFILE = None # keyfile for the certfile in PEM format. if encrypted, password will prompt on the command line
+SSL_CERTFP = None
+SSL_CIPHERS = None # custom list of available ciphers in OpenSSL cipher list format (<https://wiki.openssl.org/index.php/Manual:Ciphers(1)#CIPHER_LIST_FORMAT>)
+
```
nyuszika7h: Minor thing, but ideally sentences should start with a capital letter and there should be a period at the end of comments (not necessary for comments that are just a sentence fragment).

.gitignore
```diff
@@ -8,6 +8,9 @@
 __pycache__/
 *.py[co]
 
+# linter files
```
nyuszika7h: This is a bit of a nitpick, but can you make this "Linter" for consistency?

oyoyo/client.py
miniCruzer: The Python hashlib module does support some variable length digest hashes like blake2 and shake. There is a default for each of these however if someone wants to use a variable length digest, there still needs to be support for manual notation of the hash type.

Check for a notation, then fallback to auto-detection instead of assuming SHA256?

miniCruzer: We also have the problem of SHA3-256 using the same length as SHA2-256.

src/settings.py
```diff
@@ -6,6 +6,22 @@
 import botconfig
 
 LANGUAGE = 'en'
+
+## TLS settings
+# If SSL_CERTFP is supplied, the bot will attempt to verify that with the server. If not, then the
+# bot will attempt certificate verification, otherwise it will abort the connection.
+# You may specify the hash algorithm to use when verifying fingerprints. If unspecified, it will
+# attempt to autodetect the hash type of each fingerprint individually. (Limited to MD5, SHA1,
+# SHA224, SHA384, and SHA512).
```
skizzerz: comment is missing SHA256

src/settings.py
```diff
@@ -6,6 +6,22 @@
...
+# Multiple fingerprints may be comma separated for networks with multipleservers. Check your Python
+# version for suported algorithms.
+# Syntax: SSL_CERTFP = "[HASH-ALGO:]fingerprint1,fingerprint2,fingerprint3"
+SSL_VERIFY = True
+SSL_CERTFILE = None # Client cert file to connect with in PEM format; can also contain keyfile.
+SSL_KEYFILE = None # Keyfile for the certfile in PEM format. if encrypted, password will prompt on the command line.
+SSL_CERTFP = None
+SSL_CIPHERS = None # Custom list of available ciphers in OpenSSL cipher list format. (<https://wiki.openssl.org/index.php/Manual:Ciphers(1)#CIPHER_LIST_FORMAT>)
```
skizzerz: Depending on python version, the default when specifying None may not actually be secure. Better if we hardcode our own default here, that way we can update it to keep up with best practices instead of having every bot admin customize the list on their own.

src/settings.py
```diff
@@ -6,6 +6,22 @@
...
+
+## TLS settings
+# If SSL_CERTFP is supplied, the bot will attempt to verify that with the server. If not, then the
+# bot will attempt certificate verification, otherwise it will abort the connection.
+# You may specify the hash algorithm to use when verifying fingerprints. If unspecified, it will
+# attempt to autodetect the hash type of each fingerprint individually. (Limited to MD5, SHA1,
+# SHA224, SHA384, and SHA512).
+# Multiple fingerprints may be comma separated for networks with multipleservers. Check your Python
+# version for suported algorithms.
+# Syntax: SSL_CERTFP = "[HASH-ALGO:]fingerprint1,fingerprint2,fingerprint3"
```
skizzerz: This is dumb. Make SSL_CERTFP be a set of fingerprint strings instead of a comma-separated list. There is no need to specify hash algo explicitly as we can determine that based on string length. The default should be set() instead of None.

src/settings.py
```diff
@@ -6,6 +6,22 @@
...
+SSL_VERIFY = True
+SSL_CERTFILE = None # Client cert file to connect with in PEM format; can also contain keyfile.
+SSL_KEYFILE = None # Keyfile for the certfile in PEM format. if encrypted, password will prompt on the command line.
```
skizzerz: Prompting for key password on command line breaks automated restarts (e.g. via !restart and !update). Instead we should simply not support key files which require passwords to unlock.

Adding a config that specifies the unlock password is equivalent in security to not having a password, as anyone with the ability to read the key file and the bot source code will be able to unlock it either way. Filesystem permissions should be properly employed to secure the key instead.

src/settings.py
miniCruzer: @skizzerz Determining the hash algorithm based on length is not reliable so it needs a spot to be explicitly notated. Doesn't matter to me how it is done. It could be moved into a `SSL_CERTFP_METHOD` string and the fingerprints could be made as a set. The code will "auto-determine" common hash algorithms but if one of the more exotic hashes are used then it needs to be explicitly set.

As of Python3.6, Blake2 is supported which is a variable length hash function. Additionally,  SHA3-256 and SHA2-256 use the same length.

skizzerz: I see no point in supporting every hash algo that python supports. Pick a small handful that will be most commonly used in IRC and only support that subset. For example, I see no reason to support SHA2-224 or SHA2-384.

src/settings.py
miniCruzer: Planning to use Mozilla's "Modern" ciphersuite list.

oyoyo/client.py
```diff
@@ -174,7 +181,115 @@ def connect(self):
                         sys.exit(1)
 
             if self.use_ssl:
```
jcao219: This is a pretty big indented block.

It may be worthwhile to refactor all the TLS code into a helper function.

Then, you can call the helper function only if self.use_ssl is True, thereby eliminating the need for a large indent block.

skizzerz:
Since the original author has seemingly abandoned this, merging into a topic branch in the lykos repo so that we can push stuff to it and finish it up rather than to master.

miniCruzer:
Sorry, I have things to do that are a better use of my time than make contributions to a project where other developers refer to perfectly sane ideas as dumb. Especially given the state of the rest of the code base. 

The fingerprint string setting is based on RFC 7218.

PR #302, opened 2017-10-27
convert var.DISCONNECTED to users

skizzerz:
Also don't track disconnected users by nick anymore. But that seems less
major than fixing the docstring.

### DIFF
--- a/src/users.py
+++ b/src/users.py
@@ -91,7 +91,6 @@ def _add(cli, *, nick, ident=None, host=None, realname=None, account=None):
     This function takes up to 5 keyword-only arguments (and one positional
     argument, cli): nick, ident, host, realname and account.
     With the exception of the first one, any parameter can be omitted.
-    If a matching user already exists, a ValueError will be raised.
 
     """
 
@@ -149,6 +148,10 @@ def users():
     """Iterate over the users in the registry."""
     yield from _users
 
+def disconnected():
+    """Iterate over the users who are in-game but disconnected."""
+    yield from _ghosts
+
 def complete_match(string, users):
     matches = []
     string = lower(string)
@@ -178,10 +181,10 @@ def parse_rawnick_as_dict(rawnick, *, default=None):
 
 def _cleanup_user(evt, var, user):
     """Removes a user from our global tracking set once it has left all channels."""
-    if var.PHASE not in var.GAME_PHASES or user not in var.ALL_PLAYERS:
-        _users.discard(user)
-    elif var.PHASE in var.GAME_PHASES and user in var.ALL_PLAYERS:
+    if var.PHASE in var.GAME_PHASES and user in var.ALL_PLAYERS:
         _ghosts.add(user)
+    else:
+        _users.discard(user)
 
 def _reset(evt, var):
     """Cleans up users that left during game during game end."""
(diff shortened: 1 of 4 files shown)


src/users.py
```diff
@@ -190,13 +193,17 @@ def _reset(evt, var):
             _users.discard(user)
     _ghosts.clear()
 
+def _swap_player(evt, var, old_user, user):
+    """Marks the user as no longer being a ghost, if they are one."""
```
Vgr255: Docstrings should be imperative, i.e. "Mark the user" and not "Marks the user". Yes, other places are wrong, but that's just another reason to fix those places \o/

src/wolfgame.py
```diff
@@ -2784,39 +2781,38 @@ def reaper(cli, gameid):
                     cli.msg(chan, messages["channel_idle_warning"].format(", ".join(x)))
                 msg_targets = [p for p in to_warn_pm if p in pl]
                 mass_privmsg(cli, msg_targets, messages["player_idle_warning"].format(chan), privmsg=True)
-            for dcedplayer in list(var.DISCONNECTED.keys()):
-                acc, hostmask, timeofdc, what = var.DISCONNECTED[dcedplayer]
+            for dcedplayer, (timeofdc, what) in list(var.DISCONNECTED.items()):
```
Vgr255: No need to cast into a list if we're not mutating the variable.

src/wolfgame.py
```diff
@@ -3169,18 +3133,19 @@ def leave(var, what, user, why=None):
...
+    if user in var.ALL_PLAYERS and (what == "kick" or user in ps):
         # Prevent duplicate entry in var.ORIGINAL_ROLES
         for roleset in var.ORIGINAL_ROLES.values():
             if user.nick in roleset: # FIXME: Need to fix this once the role sets hold User instances
                 roleset.remove(user.nick)
                 roleset.add("(dced)" + user.nick) # FIXME: Need to get rid of all the (dced) hacks
                 break
 
-        var.DCED_PLAYERS[user.nick] = var.PLAYERS.pop(user.nick) # FIXME: Need to fix this once var.{DCED_}PLAYERS hold User instances
+        var.DCED_PLAYERS[user.nick] = var.PLAYERS.pop(user.nick) # FIXME: Need to fix this once var.DCED_PLAYERS hold User instances
```
Vgr255: Comment was actually accurate, it refers to both `var.PLAYERS` and `var.DCED_PLAYERS`. Admittedly, that isn't as obvious as I originally thought.

src/wolfgame.py
skizzerz: it **is** being mutated. del_player removes the user from var.DISCONNECTED

Vgr255: Oh, alright then

src/wolfgame.py
```diff
@@ -2934,27 +2903,38 @@ def fgoat(var, wrapper, message):
...
+                channels.Main.mode(("+" + hooks.Features["PREFIX"]["+"], target))
+                channels.Main.send(messages["player_return"].format(target))
+        else:
+            # this particular user doesn't exist in var.DISCONNECTED, but that doesn't
+            # mean that they aren't dced. They may have rejoined as a different nick,
+            # for example, and we want to mark them as back without requiring them to do
+            # a !swap.
+            if var.ACCOUNTS_ONLY or target.account:
+                userlist = users._get(account=target.account, allow_multiple=True) # FIXME
+            else: # match host (hopefully the ircd uses vhosts to differentiate users)
```
Vgr255: This should be `elif not var.ACCOUNTS_ONLY:`

skizzerz: Not needed. If ACCOUNTS_ONLY is set, it will always execute the if. Therefore, the else is automatically "not ACCOUNTS_ONLY" without needing to specify.

PR #303, opened 2017-11-05
Spectate changes

jacob1:
Adds more configurability to spectate, spectate is now a separate flag from revealroles and you can configure it to send a notice to the spectated chat when someone does it

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6908,9 +6908,7 @@ def can_run_restricted_cmd(user):
 
     return True
 
-@command("spectate", "fspectate", flag="a", pm=True, phases=("day", "night"))
-def fspectate(var, wrapper, message):
-    """Spectate wolfchat or deadchat."""
+def spectate_chat(var, wrapper, message, spectate_warning):
     if not can_run_restricted_cmd(wrapper.source):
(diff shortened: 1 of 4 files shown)


src/wolfgame.py
```diff
@@ -6936,19 +6934,43 @@ def fspectate(var, wrapper, message):
...
             wrapper.pm(messages["fspectate_deadchat_disabled"])
             return
         wrapper.pm(messages["fspectate_on"].format(what))
-        wrapper.pm("People in {0}: {1}".format(what, ", ".join(players)))
+        wrapper.pm("People in {0}: {1}".format(what, ", ".join([player.nick for player in players])))
+
+@command("spectate", flag="p", pm=True, phases=("day", "night"))
+def fspectate(var, wrapper, message):
+    """Spectate wolfchat or deadchat."""
+    spectate_chat(var, wrapper, message, True)
```
skizzerz: specify the True (and False in fspectate below) as a kwarg so that it is clear what that bool actually means

skizzerz: Also notifying people when spectating deadchat is dumb. Better to just not let non-admins spectate deadchat (other users can simply join it)

PR #306, opened 2017-12-08
Convert harlot to the new API

Vgr255:


src/roles/harlot.py
```diff
@@ -8,69 +8,66 @@
...
 from src import channels, users, debuglog, errlog, plog
-from src.functions import get_players, get_all_players, get_main_role
-from src.decorators import cmd, event_listener
+from src.functions import get_players, get_all_players, get_main_role, get_target
+from src.decorators import command, event_listener
 from src.messages import messages
 from src.events import Event
 
 VISITED = {}
+PASSED = set()
```
skizzerz: It'd be nice if we added type annotation comments to VISITED and PASSED. Even though we aren't using mypy (yet), it's still helpful when browsing source to see what sorts of structure things are at a glance.

src/roles/harlot.py
```diff
@@ -80,51 +77,48 @@ def on_transition_day_resolve(evt, var, victim):
...
+        if VISITED.get(harlot) in get_players(var.WOLF_ROLES) and harlot not in evt.data["dead"] and harlot not in evt.data["bitten"]:
             evt.data["message"].append(messages["harlot_visited_wolf"].format(harlot))
             evt.data["bywolves"].add(harlot)
             evt.data["onlybywolves"].add(harlot)
             evt.data["dead"].append(harlot)
 
 @event_listener("night_acted")
 def on_night_acted(evt, var, user, actor):
-    if VISITED.get(user.nick):
+    if VISITED.get(user):
```
skizzerz: can we rename these params from (evt, var, user, actor) to (evt, var, target, spy)? It is more descriptive as to what they actually represent -- spy is checking to see if target acted at night.

src/roles/harlot.py
```diff
@@ -140,6 +134,15 @@ def on_transition_night_end(evt, var):
...
     VISITED.clear()
+    PASSED.clear()
+
+@event_listener("swap_player")
+def on_swap(evt, var, old_user, user):
+    for actor, target in set(VISITED.items()):
+        if target is old_user:
+            VISITED[actor] = user
+        if actor is old_user:
+            VISITED[user] = VISITED.pop(actor)
```
skizzerz: You aren't fixing PASSED in swap_player. I'd also prefer renaming "actor" to "harlot" for extra descriptiveness :)

src/roles/succubus.py
```diff
@@ -83,11 +83,11 @@ def pass_cmd(cli, nick, chan, rest):
     chk_nightdone(cli)
 
 @event_listener("harlot_visit")
-def on_harlot_visit(evt, cli, var, nick, victim):
-    if victim in var.ROLES["succubus"]:
-        pm(cli, nick, messages["notify_succubus_target"].format(victim))
-        pm(cli, victim, messages["succubus_harlot_success"].format(nick))
-        ENTRANCED.add(nick)
+def on_harlot_visit(evt, var, actor, victim):
```
skizzerz: rename actor to harlot for proper descriptiveness

src/wolfgame.py
```diff
@@ -3517,7 +3516,7 @@ def transition_day(cli, gameid=0):
             victims.append(v)
         elif v.nick in var.ROLES["bodyguard"] and v.nick in angel.GUARDED and users._get(angel.GUARDED[v.nick]) in victims_set: # FIXME
             vappend.append(v)
-        elif v.nick in var.ROLES["harlot"] and harlot.VISITED.get(v.nick) is not None and users._get(harlot.VISITED[v.nick]) in victims_set: # FIXME
+        elif v in harlot.VISITED and harlot.VISITED[v] in victims_set:
```
skizzerz: why not "harlot.VISITED.get(v) in victims_set"?

Vgr255: Probably because I converted 1:1 and that's how it used to be.

PR #307, opened 2017-12-11
!coin, !cat, !pony

ammongit:
Move those commands to use the `@command()` decorator instead of the legacy `@cmd()` one.

### DIFF
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -6131,11 +6131,11 @@ def show_admins(cli, nick, chan, rest):
 
     channels.Main.who()
 
-@cmd("coin", pm=True)
-def coin(cli, nick, chan, rest):
+@command("coin", pm=True)
+def coin(var, wrapper, message):
     """It's a bad idea to base any decisions on this command."""
 
-    reply(cli, nick, chan, messages["coin_toss"].format(nick))
+    wrapper.send(messages["coin_toss"].format(wrapper.source.nick))
     rnd = random.random()
     # 59/29/12 split, 59+29=88
     if rnd < 0.59:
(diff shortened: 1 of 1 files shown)


src/wolfgame.py
```diff
@@ -6131,11 +6131,11 @@ def admin_endwho(event, var, target):
...
     channels.Main.who()
 
-@cmd("coin", pm=True)
-def coin(cli, nick, chan, rest):
+@command("coin", pm=True)
+def coin(var, wrapper, message):
     """It's a bad idea to base any decisions on this command."""
 
-    reply(cli, nick, chan, messages["coin_toss"].format(nick))
+    wrapper.send(messages["coin_toss"].format(wrapper.source.nick))
```
Vgr255: In `str.format` calls, you can pass in a context (e.g. `wrapper.source`) directly and it will work as intended. We try to do that everywhere that it makes sense, as we try to move away from nick-based operations :)

src/wolfgame.py
```diff
@@ -6144,14 +6144,13 @@ def coin(cli, nick, chan, rest):
...
+    wrapper.send(messages["coin_land"].format(coin))
 
-@cmd("pony", pm=True)
-def pony(cli, nick, chan, rest):
+@command("pony", "horse", pm=True)
+def pony(var, wrapper, message):
     """Toss a magical pony into the air and see what happens!"""
 
-    reply(cli, nick, chan, messages["pony_toss"].format(nick))
+    wrapper.send(messages["pony_toss"].format(wrapper.source.nick))
```
Vgr255: Ditto here

src/wolfgame.py
```diff
@@ -6161,16 +6160,15 @@ def pony(cli, nick, chan, rest):
...
 
-@cmd("cat", pm=True)
-def cat(cli, nick, chan, rest):
+@command("cat", pm=True)
+def cat(var, wrapper, message):
     """Toss a cat into the air and see what happens!"""
-    reply(cli, nick, chan, messages["cat_toss"].format(nick))
-    reply(cli, nick, chan, messages["cat_land"])
+    wrapper.send(messages["cat_toss"].format(wrapper.source.nick))
+    wrapper.send(messages["cat_land"])
```
Vgr255: Same here too, use `wrapper.source` directly. Also, the two lines can be merged into one by adding the second line into the first call with a comma, and specifying `sep="\n"` in the call.

PR #309, opened 2017-12-12
Convert var.ROLES and var.ORIGINAL_ROLES to Users

skizzerz:
This gets rid of the (dced) hack, with var.DCED_LOSERS taking its place.
Succubus was not touched, as there is already a PR which converts it,
although the succubus events in other files were touched.

Some sites were updated to be (more) Users-aware, while others replaced
it with an old deprecated get_roles() API. As more things get converted,
these will hopefully get eliminated as well. A FIXME comment was added
to all such occurrences.

### DIFF
--- a/src/gamemodes.py
+++ b/src/gamemodes.py
@@ -215,7 +215,7 @@ class VillagergameMode(GameMode):
         evt.data["transition_day"] = lambda cli, gameid=0: self.prolong_night(cli, var, gameid, transition_day)
 
     def prolong_night(self, cli, var, gameid, transition_day):
-        nspecials = len(var.ROLES["seer"] | var.ROLES["harlot"] | var.ROLES["shaman"] | var.ROLES["crazed shaman"])
+        nspecials = len(get_players(("seer", "harlot", "shaman", "crazed shaman")))
         rand = random.gauss(5, 1.5)
         if rand <= 0 and nspecials > 0:
             transition_day(cli, gameid=gameid)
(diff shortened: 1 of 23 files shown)


src/gamemodes.py
```diff
@@ -215,7 +215,7 @@ def chk_nightdone(self, evt, var):
         evt.data["transition_day"] = lambda cli, gameid=0: self.prolong_night(cli, var, gameid, transition_day)
 
     def prolong_night(self, cli, var, gameid, transition_day):
-        nspecials = len(var.ROLES["seer"] | var.ROLES["harlot"] | var.ROLES["shaman"] | var.ROLES["crazed shaman"])
+        nspecials = len(get_players(("seer", "harlot", "shaman", "crazed shaman")))
```
Vgr255: Shouldn't this be `get_all_players` instead?

src/roles/shaman.py
```diff
@@ -587,7 +587,7 @@ def on_assassinate(evt, var, killer, target, prot):
 
 @event_listener("succubus_visit")
 def on_succubus_visit(evt, cli, var, nick, victim):
-    if (SHAMANS.get(victim, (None, None))[1] in var.ROLES["succubus"] and
+    if (users._get(SHAMANS.get(victim, (None, None))[1]) in var.ROLES["succubus"] and # FIXME (psst Vgr: this probably breaks now if the shaman didn't give out totems when succ visits)
```
Vgr255: `users._get(blah, allow_none=true)`

PR #312, opened 2017-12-18
Adds 3 more "No Victim" Messages

moonheart08:
Adds the following:

    The remains of one of the village pigs are found.
    A few bones from a local squirrel are found.
    The imprint of a wolf faceplanting is found near the village square.


jcao219:
For the first two, it should be 'are found'

For the third, it should be 'is found'

;-)

moonheart08:
oh my god. how did i reverse that.

PR #313, opened 2018-01-10
Support for TLS certificate verification and client certificates redux

skizzerz:
See #301 for comments and stuff that still need to be worked on

### DIFF
--- a/src/handler.py
+++ b/src/handler.py
@@ -293,7 +293,6 @@ def connect_callback(cli):
     if botconfig.SASL_AUTHENTICATION:
         @hook("authenticate")
         def auth_plus(cli, something, plus):
-            nonlocal selected_sasl
             if plus == "+":
                 if selected_sasl == "EXTERNAL":
                     cli.send("AUTHENTICATE +")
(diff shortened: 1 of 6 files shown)


src/handler.py
```diff
@@ -272,11 +293,15 @@ def on_cap(cli, svr, mynick, cmd, *caps):
     if botconfig.SASL_AUTHENTICATION:
         @hook("authenticate")
         def auth_plus(cli, something, plus):
+            nonlocal selected_sasl
```
Vgr255: It looks like you're just using `selected_sasl`'s value and not changing it, so `nonlocal` isn't required here.

src/handler.py
```diff
@@ -287,9 +312,16 @@ def on_successful_auth(cli, blah, blahh, blahhh):
         @hook("906")
         @hook("907")
         def on_failure_auth(cli, *etc):
-            alog("Authentication failed.  Did you fill the account name "
-                 "in botconfig.USERNAME if it's different from the bot nick?")
-            cli.quit()
+            nonlocal supported_sasl, selected_sasl
```
Vgr255: Here, `supported_sasl` doesn't need to be non-local (although `selected_sasl` does).

Vgr255:
Everything code-wise looks fine to me. Someone with more knowledge in TLS/SASL than me will have to review the functional bits, however.

PR #314, opened 2018-01-17
Issue #255

### LINKED ISSUE #255: !pingif pings players who would get stasis after acknowledging their warning(s)
Now that stasis is applied after warnings are acknowledged, users who would get stasis after acknowledging their warning(s) still get pinged until they do so.

Fudster:
This fixes #255

### DIFF
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -13,7 +13,7 @@ __all__ = ["pm", "is_fake_nick", "mass_mode", "mass_privmsg", "reply",
            "relay_wolfchat_command", "irc_lower", "irc_equals", "match_hostmask",
            "is_owner", "is_admin", "plural", "singular", "list_players",
            "get_role", "get_roles", "change_role", "role_order", "break_long_message",
-           "complete_match", "complete_one_match", "get_victim", "get_nick", "InvalidModeException"]
+           "complete_match", "complete_one_match", "get_victim", "InvalidModeException"]
 # message either privmsg or notice, depending on user settings
 def pm(cli, target, message):
     if is_fake_nick(target) and botconfig.DEBUG_MODE:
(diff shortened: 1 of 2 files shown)


src/utilities.py
```diff
@@ -382,10 +382,10 @@ def get_victim(cli, nick, victim, in_chan, self_in_list=False, bot_in_list=False
...
 
 # wrapper around complete_match() used for any nick on the channel
-def get_nick(cli, nick):
-    ul = [x for x in var.USERS]
-    ull = [x.lower() for x in var.USERS]
-    lnick = complete_match(nick.lower(), ull)
+def get_nick(nick):
+    ul = list(var.USERS.keys())
+    ull = [u.lower() for u in ul]
+    lnick = complete_one_match(nick.lower(), ull)
```
skizzerz: get_nick is unused. Remove it entirely (remember to remove it from __all__ at the top of the file as well).

src/wolfgame.py
```diff
@@ -673,12 +673,19 @@ def join_timer_handler(var):
             for num in var.PING_IF_NUMS_ACCS:
                 if num <= len(pl):
                     for acc in var.PING_IF_NUMS_ACCS[num]:
+                        temp = users._get(get_nick(acc))
```
skizzerz: 1. call the variable something other than temp (perhaps "user")
2. eliminate the get_nick call: user = users._get(account=acc)

src/wolfgame.py
```diff
@@ -673,12 +673,19 @@ def join_timer_handler(var):
...
+                        if db.has_unacknowledged_warnings(temp.account, temp.rawnick):
+                            continue
                         chk_acc.add(users.lower(acc))
 
         if not var.ACCOUNTS_ONLY:
             for num in var.PING_IF_NUMS:
                 if num <= len(pl):
                     for hostmask in var.PING_IF_NUMS[num]:
+                        host = hostmask[hostmask.find("@") + 1:]
+                        temp = users._get(host=host)
```
skizzerz: has_unacknowledged_warnings's hostmask param should probably take an ident@host, not a nick!ident@host. You can eliminate both of these above lines and just pass hostmask to it directly.

(I know there are things that pass the full mask in, but I think those things are wrong. I'll address them outside of this PR)

PR #317, opened 2018-04-16
Convert chk_decision

Vgr255:
This PR converts a large part of the code to use the better User handling. The core idea is that `chk_decision` and all of its branching events (recursively) are converted (and, in some cases, its parents). Also being converted are `transition_night`, `transition_day`, and `chk_nightdone`. I've also taken the liberty of converting `var.GUNNERS` and splitting shamans into four files; a helper and each shaman role in a separate file. There are other some small changes a bit everywhere. Tested thanks to Iciloo and MoonyTheDwarf, ready to merge after a review!


src/roles/_shaman_helper.py
```diff
@@ -0,0 +1,423 @@
...
+        user.send(messages["totem_narcolepsy"])
+        evt.prevent_default = True
+
+@event_listener("assassinate")
+def on_assassinate(evt, var, killer, target, prot):
+    if prot == "totem":
+        var.ACTIVE_PROTECTIONS[target.nick].remove("totem")
+        evt.prevent_default = True
+        evt.stop_processing = True
+        channels.Main.send(messages[evt.params.message_prefix + "totem"].format(killer, target))
```
skizzerz: Should figure out what message_prefixes we could be passed and add a comment above this listing the full message keys for each combination. That way when we see a message in en.json we can figure out where it's being used by doing a grep for the key.

src/roles/_shaman_helper.py
```diff
@@ -0,0 +1,423 @@
...
+#
+# To add new shaman roles in your custom roles/whatever.py file:
+# 1. Expand var.TOTEM_ORDER and upate var.TOTEM_CHANCES to account for the new width
+# 2. Add the role to var.ROLE_GUIDE
+# 3. Add the role to whatever other holding vars are necessary based on what it does
+# 4. Implement custom events if the role does anything else beyond giving totems.
+#
+# Modifying this file to add new totems or new shaman roles is generally never required
+
+def _get_target(var, wrapper, message, lastgiven):
```
skizzerz: Missing docstring (as is _give_totem). Name shouldn't start with _ as this is meant to be imported and used by shaman roles.

src/roles/_shaman_helper.py
```diff
@@ -0,0 +1,423 @@
...
+havetotem = []              # type: List[users.User]
+brokentotem = set()         # type: Set[users.User]
+
+# To add new totem types in your custom roles/whatever.py file:
+# 1. Add a key to var.TOTEM_CHANCES with the totem name
+# 2. Add a message totemname_totem to your custom messages.json describing
+#    the totem (this is displayed at night if !simple is off)
+# 3. Add events as necessary to implement the totem's functionality
+#
+# XXX The paragraph below is obsolete, needs to be updated
```
skizzerz: Still seems accurate enough, leave as-is.

src/roles/crazed_shaman.py
```diff
@@ -0,0 +1,180 @@
...
+
+@command("give", "totem", chan=False, pm=True, playing=True, silenced=True, phases=("night",), roles=("crazed shaman",))
+def crazed_shaman_totem(var, wrapper, message, prefix="You"):
+    """Give a random totem to a player."""
+
+    target = _get_target(var, wrapper, message, LASTGIVEN)
+    if not target:
+        return
+
+    _give_totem(var, wrapper, target, prefix, set(), "crazed shaman", "", SHAMANS)
```
skizzerz: I have a couple of issues with the aesthetics of this call.
1. I don't like how it mutates SHAMANS. Can we have give_totem return the thing we use to add to SHAMANS instead? e.g. SHAMANS[wrapper.source] = give_totem(var, wrapper, target, ...)
2. I don't like passing set() and "" as positional arguments -- it's very non-obvious what those are for. If it's necessary to pass them at all rather than them just taking default values from give_totem, can we pass them as kwargs?

src/roles/crazed_shaman.py
```diff
@@ -0,0 +1,180 @@
...
+from src.events import Event
+
+from src.roles._shaman_helper import havetotem, _get_target, _give_totem, _apply_totem
+
+TOTEMS = UserDict()         # type: Dict[users.User, str]
+LASTGIVEN = UserDict()      # type: Dict[users.User, users.User]
+SHAMANS = UserDict()        # type: Dict[users.User, List[users.User]]
+
+@command("give", "totem", chan=False, pm=True, playing=True, silenced=True, phases=("night",), roles=("crazed shaman",))
+def crazed_shaman_totem(var, wrapper, message, prefix="You"):
```
skizzerz: get rid of prefix as a kwarg, and kill the hardcoded English.

src/roles/crazed_shaman.py
```diff
@@ -0,0 +1,180 @@
...
+from src.containers import UserList, UserSet, UserDict, DefaultUserDict
+from src.dispatcher import MessageDispatcher
+from src.messages import messages
+from src.events import Event
+
+from src.roles._shaman_helper import havetotem, _get_target, _give_totem, _apply_totem
+
+TOTEMS = UserDict()         # type: Dict[users.User, str]
+LASTGIVEN = UserDict()      # type: Dict[users.User, users.User]
+SHAMANS = UserDict()        # type: Dict[users.User, List[users.User]]
```
skizzerz: Prefer we add helper methods for this:
TOTEMS, LASTGIVEN, SHAMANS = setup_totem_vars()

The purpose of setup_totem_vars() is to create the three UserDicts and return them (as a tuple), but also to wire up events in the shared shaman code to do things like reset, begin_day, etc. Anywhere where the events are 100% the same across every shaman role (notably, transition_night_end is not included since the totem assignment and role messages code varies). This will reduce a lot of boilerplate.

src/roles/_shaman_helper.py
```diff
@@ -0,0 +1,423 @@
...
+    elif totem == "misdirection":
+        MISDIRECTION.add(victim)
+    elif totem == "deceit":
+        DECEIT.add(victim)
+    # other totem types possibly handled in an earlier event,
+    # as such there is no else: clause here
+
+@event_listener("see", priority=10)
+def on_see(evt, var, nick, victim):
+    if (users._get(victim) in DECEIT) ^ (users._get(nick) in DECEIT): # FIXME
```
skizzerz: You updated the see event to pass users... did you not actually test all of the totems?

Vgr255:
Will check when I'm home. I seem to recall asking for every totem to be tested, but I didn't do the testing.

src/wolfgame.py
```diff
@@ -1720,73 +1730,68 @@ def stats(cli, nick, chan, rest):
...
+            "not_lynching": not_lynching,
+            "votelist": votelist,
+            "numvotes": {}, # filled as part of a priority 1 event
+            "weights": {}, # filled as part of a priority 1 event
+            "transition_night": transition_night
+            }, voters=pl, timeout=True)
+        if not event.dispatch(var, None):
+            return
+        not_lynching = event.data["not_lynching"]
+        votelist = event.data["votelist"]
```
skizzerz: Does this work in python? Reassigning the thing you're context-managing inside of the context manager seems like something that shouldn't happen. (I know in C# it throws an exception if you try, idk what python's behavior is)

src/wolfgame.py
```diff
@@ -1795,108 +1800,108 @@ def fday(cli, nick, chan, rest):
     if var.PHASE != "night":
         cli.notice(nick, messages["not_nighttime"])
     else:
-        transition_day(cli)
+        transition_day()
 
-# Specify force = "nick" to force nick to be lynched
-def chk_decision(cli, force="", end_game=True, deadlist=[]):
+# Specify force = user to force user to be lynched
+def chk_decision(force=None, end_game=True, deadlist=UserList()):
```
skizzerz: The deadlist kwarg is a ticking time bomb waiting to happen (it was an issue back when the default was [] as well). Should default to None instead and figure out what to do in that case. Right now you create a new UserList that will never under any circumstance be cleared out should users get added to it (can users even be added to it? or do we make a copy right away and only operate on the copy?).

src/wolfgame.py
```diff
@@ -1795,108 +1800,108 @@ def fday(cli, nick, chan, rest):
...
-                        lmsg = random.choice(messages["lynch_no_reveal"]).format(votee)
-                    cli.msg(botconfig.CHANNEL, lmsg)
-                    better_deadlist = [users._get(p) for p in deadlist] # FIXME -- convert chk_decision_lynch to be user-aware
-                    if not del_player(users._get(votee), killer_role="villager", deadlist=better_deadlist, end_game=end_game): # FIXME
-                        return
-                do_night_transision = True
-                break
-        if do_night_transision:
-            event.data["transition_night"](cli)
+                        del votelist[users.Bot]
```
skizzerz: The above block could probably be made a high-priority event listener in gamemodes.py (like priority 20) instead of being in wolfgame.py since it's gamemode-specific code. Listener would need to be added to both the default mode as well as the villagergame mode.

src/wolfgame.py
```diff
@@ -3383,23 +3371,27 @@ def transition_day(cli, gameid=0):
     # 6 = rearranging victim list (ensure bodyguard/harlot messages plays),
     #     fixing killers dict priority again (in case step 4 or 5 added to it)
     # Actually killing off the victims happens in transition_day_resolve
+    # We set the variables here first; listeners should mutate, not replace
+    # We don't need to use User containers here, as these don't persist long enough
+    # This removes the burden of having to clear them at the end or should an error happen
```
skizzerz: Comment block seems unnecessary, and getting rid of the ability to replace data dict entries if they're containers is something I probably think we should do in general (doesn't need to be part of this PR though).

That said, you missed a bunch of places using revt.data["blah"] below which was made under the assumption that replacement is possible -- can replaces those with the vars directly.

src/wolfgame.py
```diff
@@ -4126,73 +4102,62 @@ def check_exchange(cli, actor, nick):
...
-    if victim == wrapper.source.nick:
-        wrapper.pm(messages["gunner_target_self"])
+
+    target = get_target(var, wrapper, re.split(" +", message)[0], not_self_message="gunner_target_self")
+    if not target:
         return
+
     # get actual victim
-    victim = choose_target(wrapper.source.nick, victim)
+    target = users._get(choose_target(wrapper.source.nick, target.nick)) # FIXME
```
skizzerz: Use get_target() here and fire a targeted_command event (misdirection should apply for sure, unsure about exchange but that'd be funny too)

src/wolfgame.py
```diff
@@ -6875,15 +6857,50 @@ def fgame(cli, nick, chan, rest):
...
     else:
-        cli.notice(nick, fgame.__doc__())
+        wrapper.pm(fgame.__doc__())
+
+@command("frole", flag="d", phases=("join",))
+def frole(var, wrapper, message):
+    """Force a player into a certain role."""
+    pl = get_players()
+    if wrapper.source not in pl and not wrapper.source.is_admin():
+        return
```
skizzerz: Why this check? It's a debug mode only command (or will be in the nearish future -- I plan on handling all commands with the "d" flag as debug-mode only and get rid of the debug mode check at the bottom of wolfgame.py) and seems annoying for testers when we are in debug mode because it forces the tester to !join.

src/wolfgame.py
```diff
@@ -6875,15 +6857,50 @@ def fgame(cli, nick, chan, rest):
...
+
+    to_force = {}
+
```
skizzerz: I don't like how you made this work. Requiring underscores in role names is not something we do anywhere else (e.g. !fgame roles). Make some sort of punctuation like comma required, and split on that instead of replacing all of that with spaces. Then, we can correctly handle spaces in role names.

src/roles/_shaman_helper.py
skizzerz: Right, I misread a thing. This is fine and doesn't need to be changed.

src/wolfgame.py
jacob1: Yeah this was annoying me too. I'd rather type commas than underscores

src/roles/_shaman_helper.py
```diff
@@ -0,0 +1,423 @@
...
+
+    return target
+
+def _give_totem(var, wrapper, target, prefix, tags, role, msg, shamans):
+
+    orig_target = target
+    orig_role = get_main_role(orig_target)
+
+    evt = Event("targeted_command", {"target": target, "misdirection": True, "exchange": True},
+                action = "give a totem{0} to".format(msg))
```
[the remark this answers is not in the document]
Vgr255: I literally copied the code that used to be in `shaman.py`. I fixed the spaces but the hardcoded English is something to be fixed outside of the PR.

src/wolfgame.py
Vgr255: It works no problem (was tested), and is intended as it's the variable we should be clearing at the end. Though to be honest, I probably shouldn't re-assign it and just let events mutate.

src/wolfgame.py
Vgr255: Fair point, I'll look into it.

src/wolfgame.py
Vgr255: I'll let @jacob1 fix this command once the PR is merged, so won't touch it further.

PR #320, opened 2018-04-24
Split time lord

ammongit:


src/roles/timelord.py
```diff
@@ -0,0 +1,68 @@
...
+        if timer_name in var.TIMERS:
+            timer = var.TIMERS[timer_name][0]
+            if timer.isAlive():
+                timer.cancel()
+                t = threading.Timer(time_warn, hurry_up, [cli, phase_id, False])
+                var.TIMERS[timer_name] = (t, time.time(), time_warn)
+                t.daemon = True
+                t.start()
+
+    debuglog("{0} (time lord) TRIGGER".format(player.nick))
```
Vgr255: Only put `player` in here, no need for `player.nick` (for consistency).

src/roles/timelord.py
```diff
@@ -0,0 +1,69 @@
...
+        time_warn = var.DAY_TIME_WARN
+        phase_id = "DAY_ID"
+        timer_name = "day_warn"
+    elif var.GAMEPHASE == "night":
+        time_limit = var.NIGHT_TIME_LIMIT
+        time_warn = var.NIGHT_TIME_WARN
+        phase_id = "NIGHT_ID"
+        timer_name = "night_warn"
+    else:
+        return
```
Vgr255: I think we might want to raise here, as if `var.GAMEPHASE` is neither day or night, then I have no clue what it *is* and we might want to error. Alternatively, don't do anything, and just let the code below break if it gets to it.

ammongit: Alright. I wasn't sure if those phases were exhaustive.

PR #322, opened 2018-04-25
Allow pipers to pass or change targets

ammongit:
Also adds additional details to `!revealroles` that shows how is going to be charmed by the next day.


src/roles/piper.py
```diff
@@ -85,6 +85,18 @@ def charm(var, wrapper, message):
         debuglog("{0} (piper) CHARM {1} ({2})".format(wrapper.source, target1, get_main_role(target1)))
         wrapper.send(messages["charm_success"].format(orig1))
 
+@command("pass", chan=False, pm=True, playing=True, silenced=True, phases=("night",), roles=("piper",))
+def pass_cmd(var, wrapper, message):
+    """Do not charm anyone tonight."""
+
+    debuglog("{0} (piper) CHARM NOBODY".format(wrapper.source))
+    wrapper.send(messages["piper_pass"])
```
Vgr255: Message and logging should go after the actual logic.

src/roles/piper.py
```diff
@@ -200,6 +212,17 @@ def on_reset(evt, var):
...
+    if TOBECHARMED:
+        if len(get_players(("piper",))) == 1:
+            message = messages["piper_revealroles_tocharm"]
+        else:
+            message = messages["piper_revealroles_tocharm_user"]
+
+        for piper, to_charm in TOBECHARMED.items():
+            if to_charm:
+                nicks = ", ".join(p.nick for p in to_charm)
+                evt.data["output"].append(message.format(nicks, piper.nick))
```
Vgr255: I would put all of the `TOBECHARMED` logic under a `revealroles_role` listener instead, and have it a per-user thing. For example, `piper: aismallard (charming Vgr, woffle) | ... | charmed players: jacob1, ilbelkyr`.

src/roles/piper.py
```diff
@@ -160,7 +183,7 @@ def on_transition_day_begin(evt, var):
```
Vgr255: Should do `PASSED.clear()` here (at the end of `transition_day_begin`).

src/roles/piper.py
```diff
@@ -196,10 +219,19 @@ def on_acted(evt, var, target, spy):
 def on_reset(evt, var):
     CHARMED.clear()
     TOBECHARMED.clear()
+    PASSED.clear()
```
Vgr255: Need to handle `PASSED` (and `TOBECHARMED` too, apparently) in the `exchange_roles` listener. Make sure to test as well.

ammongit:
Also changed some places to use the `del USERDICT[:user:]` syntax.

src/roles/hunter.py
```diff
@@ -43,20 +43,20 @@ def hunter_kill(var, wrapper, message):
 @command("retract", "r", chan=False, pm=True, playing=True, phases=("night",), roles=("hunter",))
 def hunter_retract(var, wrapper, message):
     """Removes a hunter's kill selection."""
-    if wrapper.source not in KILLS and wrapper.source not in PASSED:
-        return
```
Vgr255: This shouldn't be removed. It's there to prevent a hunter who hasn't killed or passed to retract.

src/roles/vengefulghost.py
```diff
@@ -49,16 +49,18 @@ def vg_kill(var, wrapper, message):
...
     if wrapper.source not in GHOSTS:
         return
-    if wrapper.source in KILLS:
-        del KILLS[wrapper.source]
-        wrapper.pm(messages["retracted_kill"])
+
+    del KILLS[:wrapper.source:]
+
+    wrapper.pm(messages["retracted_kill"])
+    debuglog("{0} (vengeful ghost) RETRACT".format(wrapper.source))
```
Vgr255: Need to keep the check under `if wrapper.source in KILLS` because you don't want to send a message for no reason.

src/roles/hunter.py
```diff
@@ -45,18 +45,22 @@ def hunter_retract(var, wrapper, message):
     """Removes a hunter's kill selection."""
     if wrapper.source not in KILLS and wrapper.source not in PASSED:
         return
-    KILLS.pop(wrapper.source, None)
+
+    del KILLS[wrapper.source]
```
Vgr255: Need the :: del here

src/roles/hunter.py
```diff
@@ -45,18 +45,22 @@ def hunter_retract(var, wrapper, message):
...
 
 @command("pass", chan=False, pm=True, playing=True, silenced=True, phases=("night",), roles=("hunter",))
 def hunter_pass(var, wrapper, message):
     """Do not use hunter's once-per-game kill tonight."""
     if wrapper.source in HUNTERS and wrapper.source not in KILLS:
         wrapper.pm(messages["hunter_already_killed"])
         return
-    KILLS.pop(wrapper.source, None)
+
+    del KILLS[wrapper.source]
```
Vgr255: Also needs the :: del here

PR #328, opened 2018-04-30
Add back detective to 4p mudkip

ammongit:
Resolves the "investigator is useless at 4p" problem, one solution is to make the investigator a detective at 4p only, like in original mudkip.


skizzerz:
The change was made for balance purposes. Village winrate was too high across the board including in 4p. It is not going to be reverted.

PR #335, opened 2018-05-02
Add centralized error reporting (off by default)

skizzerz:
If enabled, sends error details over to us so that we can look at them.
Includes a lot more details than the standard pastebin does, including
full IRC logs for the game as well as various game state vars that lets
us see hidden stuff more easily.

Still a work in progress, could probably be expanded to be more useful.
Compressing the data we stream over is probably not a bad idea either...

### DIFF
--- a/src/decorators.py
+++ b/src/decorators.py
@@ -121,9 +121,25 @@ class print_traceback:
 
         variables[1] = _local.handler.traceback
 
-        if not botconfig.PASTEBIN_ERRORS or channels.Main is not channels.Dev:
+        if var.REPORT_ERRORS:
+            api_url = "https://werewolf.chat/devsite/report.php"
+            api_body = logger.get_replay(source)
+            api_body["traceback"] = variables
+
+            data = None
+            with _local.handler:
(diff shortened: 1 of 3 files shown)


src/decorators.py
```diff
@@ -121,9 +121,25 @@ def __exit__(self, exc_type, exc_value, tb):
 
         variables[1] = _local.handler.traceback
 
-        if not botconfig.PASTEBIN_ERRORS or channels.Main is not channels.Dev:
+        if var.REPORT_ERRORS:
+            api_url = "https://werewolf.chat/devsite/report.php"
+            api_body = logger.get_replay(source)
```
Vgr255: This should be inside a `with _local.handler:` block. It's unlikely that it does fail, but if it does this may potentially crash the bot and wipe out any log. I'm not sure what we should do if it does fail, but probably continue without any form of replay.

skizzerz:
Outdated and don't really have time to revisit this. We now have centralized reporting of error tracebacks, sans replays. I'd still like to add replays eventually.

PR #338, opened 2018-05-27
Convert wild child

ammongit:


src/roles/wildchild.py
```diff
@@ -1,78 +1,69 @@
-import random
 import re
+import random
+import itertools
+import math
+from collections import defaultdict
 
-import src.settings as var
+import botconfig
```
skizzerz: don't import botconfig in role files. It is never needed. If you need botconfig in a role file, you're doing something wrong so change what you're doing so it isn't needed.

src/roles/wildchild.py
```diff
@@ -81,45 +72,42 @@ def on_myrole(evt, var, user):
...
-                if wolf in get_all_players(("cursed villager",)):
-                    cursed = "cursed "
-                new.append("\u0002{0}\u0002 ({1}{2})".format(wolf, cursed, wolfroles))
-
-            if new:
-                child.send(messages["wolves_list"].format(", ".join(new)))
-            else:
-                child.send(messages["no_other_wolves"])
+
+            wolves = get_all_players(wcroles)
```
skizzerz: get_players is correct here. People assigned wolfteam roles as secondary are not actual wolves and are thus not meant to be notified.

src/roles/wildchild.py
```diff
@@ -131,29 +119,27 @@ def on_chk_nightdone(evt, var):
...
-                pl.remove(child)
-                if pl:
-                    target = random.choice(pl)
-                    IDOLS[child.nick] = target.nick
-                    child.send(messages["wild_child_random_idol"].format(target))
+            if child not in IDOLS:
+                players = get_players()
+                players.remove(child)
+                if players:
+                    random_idol(child, players)
```
skizzerz: keep this inline. Don't split something out into a helper method unless it will actually be called from multiple places.

src/roles/wildchild.py
```diff
@@ -1,78 +1,69 @@
...
 from src.events import Event
 
-WILD_CHILDREN = set()
-IDOLS = {}
+WILD_CHILDREN = UserSet()
+IDOLS = UserDict()
 
-@cmd("choose", chan=False, pm=True, playing=True, phases=("night",), roles=("wild child",))
-def choose_idol(cli, nick, chan, rest):
+@command("choose", chan=False, pm=True, playing=True, silenced=False, phases=("night",), roles=("wild child",))
```
skizzerz: silenced=False is the default, so this addition isn't really needed.

src/roles/wildchild.py
ammongit: I split it off because the indentation level was getting really deep and making it hard to read.

src/roles/wildchild.py
ammongit: Alright.

src/roles/wildchild.py
ammongit: I just copied the imports from `roles/_skel.py`. I don't use it anywhere.

src/roles/wildchild.py
ammongit: Ok.

src/roles/wildchild.py
```diff
@@ -85,41 +69,41 @@ def on_del_player(evt, var, user, mainrole, allroles, death_triggers):
...
-                    cursed = "cursed "
-                new.append("\u0002{0}\u0002 ({1}{2})".format(wolf, cursed, wolfroles))
-
-            if new:
-                child.send(messages["wolves_list"].format(", ".join(new)))
-            else:
-                child.send(messages["no_other_wolves"])
+
+            wolves = get_players(wcroles)
+            if wolves:
```
Vgr255: Forgot to remove `child` from `wolves` before this line.

src/roles/wildchild.py
```diff
@@ -131,29 +115,31 @@ def on_chk_nightdone(evt, var):
...
-        if user.nick in IDOLS:
-            evt.data["special_case"].append("picked {0} as idol".format(IDOLS[user.nick]))
+        if user in IDOLS:
+            evt.data["special_case"].append(messages["wild_child_revealroles_picked"].format(IDOLS[user]))
         else:
-            evt.data["special_case"].append("no idol picked yet")
+            evt.data["special_case"].append(messages["wild_child_revealroles_no_idol"])
 
 @event_listener("get_reveal_role")
 def on_get_reveal_role(evt, var, user):
```
Vgr255: Forgot to update this listener.

src/roles/wildchild.py
```diff
@@ -1,78 +1,62 @@
...
+        WILD_CHILDREN.add(actor)
     elif target_role == "wild child" and actor_role != "wild child":
         # if they're both wild children, already swapped idols above
-        IDOLS[actor.nick] = IDOLS.pop(target.nick)
-        evt.data["actor_messages"].append(messages["wild_child_idol"].format(IDOLS[actor.nick]))
+        IDOLS[actor] = IDOLS.pop(target)
+        evt.data["actor_messages"].append(messages["wild_child_idol"].format(IDOLS[actor]))
 
 @event_listener("myrole")
 def on_myrole(evt, var, user):
```
Vgr255: Missed this listener.

src/roles/wildchild.py
```diff
@@ -1,78 +1,62 @@
...
-    IDOLS[nick] = victim
-    pm(cli, nick, messages["wild_child_success"].format(victim))
-
-    debuglog("{0} (wild child) IDOLIZE: {1} ({2})".format(nick, victim, get_role(victim)))
+    IDOLS[wrapper.source] = idol
+    wrapper.send(messages["wild_child_success"].format(idol))
+    debuglog("{0} (wild child) IDOLIZE: {1} ({2})".format(wrapper.source, idol, get_main_role(idol)))
 
 @event_listener("see")
 def on_see(evt, var, seer, target):
```
Vgr255: Missed this listener.

PR #339, opened 2018-05-31
Create alias for wolf kill

ammongit:
An important quality-of-life change that will make playing a wolf role more immersive.


moonheart08:
this is the best commit in ages. please add it my game experience is broken without it

skizzerz:
Per discussion in #lykos about related commands

PR #341, opened 2018-06-24
Add total winrate output to !p

electricalabrocoma:
Mostly motivated by https://werewolf.chat/leaderboard/ listing only people with 200+ games

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -352,9 +352,13 @@ def get_player_totals(acc, hostmask):
     totals = []
     for row in c:
         tmp[row[0]] = row[1]
-    c.execute("""SELECT TOTAL(gp.team_win OR gp.indiv_win)/COUNT(1)
+    c.execute("""SELECT SUM(gp.team_win | gp.indiv_win)
                         FROM game_player gp
-                        WHERE gp.player = ?""", (peid,))
+                        JOIN player pl
+                          ON gp.player = pl.id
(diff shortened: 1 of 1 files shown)


src/db.py
```diff
@@ -352,13 +352,17 @@ def get_player_totals(acc, hostmask):
     totals = []
     for row in c:
         tmp[row[0]] = row[1]
+    c.execute("""SELECT TOTAL(gp.team_win OR gp.indiv_win)/COUNT(1)
+                        FROM game_player gp
+                        WHERE gp.player = ?""", (peid,))
+    winrate = c.fetchone()[0]
```
skizzerz: game_player is keyed on player id, not person id. This query is entirely wrong because of this. You’ll still want to pass in person id (peid) as the param, but you’ll need to perform player coalescing by joining to player and person like all the other queries in this file.

Use SUM() instead of TOTAL(). I’m not a fan of the OR either; find a more standard way to express this as OR in standard SQL is meant for boolean contexts only. The fact that it “works” in SQLite is an implementation detail. Our SQL statements should use as few dbms-specific tricks as possible in order to make them more understandable by a wide variety of people and to make porting to other dbmses easier in the future. A bitwise or would likely be appropriate here as opposed to logical or.

electricalabrocoma: fixed.

PR #344, opened 2018-07-01
Add !rolestats command

random-nick:
This might me a little confusing, because it calculates the number of players that won (similar to !playerstats) instead of the number of times a team won (like !gamestats). This way it gives the proper winrate of a player playing that role, but gives an inflated number of total victories and games in cases where there are multiples of a role in the same game. Maybe it would be better to hide the actual numbers and show only the winrates in order to avoid confusion.


src/utilities.py
```diff
@@ -383,6 +385,26 @@ def get_victim(cli, nick, victim, in_chan, self_in_list=False, bot_in_list=False
         return
     return pl[pll.index(tempvictims.pop())] #convert back to normal casing
 
+#completes a partial role
+def complete_role(cli, nick, chan, role):
```
skizzerz: utilities.py is deprecated; new things should go in functions.py and be user-aware. A nick completion function is already in functions.py as get_target() that you can base this function on.

src/wolfgame.py
```diff
@@ -5716,32 +5718,59 @@ def player_stats(cli, nick, chan, rest):
...
+        if role:
+            reply(cli, nick, chan, db.get_player_stats(acc, hostmask, role))
+        
 @cmd("mystats", "m", pm=True)
 def my_stats(cli, nick, chan, rest):
     """Get your own stats."""
     rest = rest.split()
     player_stats.func(cli, nick, chan, " ".join([nick] + rest))
 
+@cmd("rolestats", "rstats", pm=True)
```
skizzerz: Please use the new command decorator rather than cmd. All of the following stuff will need to be redone to accommodate this as well since it will be based on user objects instead of nick strings.

src/db.py
```diff
@@ -463,6 +463,72 @@ def get_game_totals(mode):
...
+                   SUM(gp.team_win) AS team,
+                   SUM(gp.indiv_win) AS indiv,
+                   SUM(gp.team_win OR gp.indiv_win) AS overall,
+                   COUNT(1) AS total
+                 FROM game g
+                 JOIN game_player gp
+                   ON gp.game = g.id
+                 JOIN game_player_role gpr
+                   ON gpr.game_player = gp.id
+                 WHERE role = ?""", (role,))
```
skizzerz: missing GROUP BY clause to actually perform the aggregation.

src/db.py
```diff
@@ -463,6 +463,72 @@ def get_game_totals(mode):
...
+                   SUM(gp.indiv_win) AS indiv,
+                   SUM(gp.team_win OR gp.indiv_win) AS overall,
+                   COUNT(1) AS total
+                 FROM game g
+                 JOIN game_player gp
+                   ON gp.game = g.id
+                 JOIN game_player_role gpr
+                   ON gpr.game_player = gp.id
+                 WHERE role = ?
+                   AND g.gamemode = ?""", (role, mode))
```
skizzerz: missing GROUP BY clause to actually perform the aggregation

PR #348, opened 2018-07-06
Redo all the things

Vgr255:
Opening PR so reviewing/commenting is easier. Don't merge yet.


skizzerz:
I would rather keep "neutral" alongside "win stealer" -- it makes the code SO MUCH NICER than if we separate them into two separate categories.

Neutral does not mean "teamless", it just means "not on wolf team and not on village team". Win stealer roles are therefore neutral by that definition, and it makes roles which interact with neutrals as well as the stop_game code to determine winners a lot cleaner if all neutrals (including win stealers) are lumped together in that category.

My design for the categories is (in general) as follows:
* Exactly one of wolfteam, village, neutral is required. Any more or less than one is an error. This determines overall team affiliation insofar as stop_game is concerned. Wolfteam wins with wolves, village wins with village, and neutral is responsible for implementing an event to determine its own win conditions.
* It is an error to have wolfchat or wolf without it also being wolfteam. It is possible to have wolf without wolfchat (sorcerer, hag, warlock, traitor, etc.) or wolfchat without wolf (a theoretical "lone wolf" role) without issue.
* All other categories are decorative (as in, not used by core gameplay mechanics, but may be used by role-specific stuff). This includes "win stealer" -- stop_game doesn't care about that since the definition of "neutral" means the win stealer is implementing its own win handling anyway.

Unless you have any specific objections, I'll be reverting that change in due course.

Vgr255:
Hmm, makes sense actually! My motivation behind the change was in the case of some role which cared about which roles win as a team vs which roles win individually (while being a win stealer). I didn't realize that "neutral" had the implication of being a core game mechanic.

However, I still think that there should be a way to distinguish roles that win as a team such as monsters, pipers, succubi etc., and roles that don't, such as fool, but there probably is a better way to do it than hijacking the "neutral" tag. Maybe a "tertiary team" tag or something alike? That would imply the role is responsible for handling that kind of stuff. Don't bother reverting, I'll do it in my next push (won't add anything else for now).

Open questions: Should a set containing more than one of village, wolfteam, or neutral be considered an error? Should also all win stealers be neutral and not any other team?

Additionally and unrelatedly, is it worth trying to figure out a common set that encapsulates all roles that amnesiac can't turn into? There is some logic to it, we just haven't laid it out - doing so would maybe help us figure out a tag for that.

Vgr255:
Possible new tags to add (names placeholders, feel free to change):
`default`: roles with no ability or anything. includes villager and cultist, and maybe drunk
`beginning`: roles which only act at the beginning of the game. includes matchmaker, clone, wild child, and possibly minion
`longterm`: roles which need time to properly setup whatever condition they have. includes piper and dullahan, and maybe doctor (although it might fit more in beginning)
`wolf-like` (`team-switcher`?): roles which are not wolves but may become so during game. includes lycan and wild child, and also traitor possibly
`unsafe`: roles for which their game winning chance is greatly altered (either in a good or bad way) by being publicly safe. includes fool and jester

Actual applications of the new tags:

- Amnesiac blacklist: roles which cannot be amnesiac have one of the following tags: {`default`, `win stealer`, `beginning`, `longterm`}
  - May be able to merge `beginning` and `longterm` into one I guess
- Template restrictions, also known as secondary roles:
  - Cursed can't be anyone whose main role has any one of {`wolf`, `cursed`, `unsafe`}
    - Outliers: seer, oracle, priest
  - Gunner/sharpshooter can't be anyone whose main role has any of {`wolfteam`, `wolf-like`, `unsafe`}
    - Outliers: priest
  - Mayor can't be anyone whose main role has any of {`unsafe`, `win stealer`}
  - Assassin can't be anyone whose main role has any of {`wolf`, `wolf-like`}
    - Outliers: seer, oracle, augur, harlot, detective, bodyguard, guardian angel, priest
    - Fallen angel *can* be assassin, but that should be FA's job to handle
  - Blessed villager would be better implemented as a whitelist consisting only of {`villager`}

Vgr255:
The `safe` category doesn't make much sense right now... Seer seeing the player as the role they are should be the default behaviour, and that should be altered by other categories, such as `innocent`, `cursed`, etc. I propose we remove the category - or, rather, re-purpose the name to mean "roles that have a good standing with village", using it for the restrictions for assassin instead.

Speaking of secondary roles restrictions, I've come up with a simple syntax for that. It would look like this:

```
    self.SECONDARY_ROLES = {
        "cursed villager": ["*", "-cursed", "-wolf", "-unsafe"],
        "gunner": ["village", "neutral", "hidden", "-unsafe", "-team-switcher"],
        "sharpshooter": ["village", "neutral", "hidden", "-unsafe", "-team-switcher"],
        "mayor": ["*", "-unsafe", "-win stealer"],
        "assassin": [...], # haven't quite figured that one out yet (see also above)
        "blessed villager": ["default", "-wolfteam"],
    }
```
where `"*"` means to add all roles to the whitelist. Any tag adds all relevant roles to the whitelist, and a tag preceded by a `-` removes all relevant roles from the whitelist. It's a whitelist in essence, but can be used as a blacklist with the use of minus signs. All categories are evaluated in order, so you could do some interesting stuff with it.

I have a working proof-of-concept on my bot if you're interested.

skizzerz:
villager and cultist are special in that have some hardcoded stuff in core gameplay; you can tag them if you want but nothing else should additionally be in that category (drunk doesn't fit, for instance). "default" seems like a bad choice, since that implies that they are the default roles (when indeed any role can be a default role now with this PR). Amnesiac hardcoding villager and cultist seems fine to me, given that.

beginning: part of this PR is eliminating that entirely. I see value in a tag that indicates the role is required to act the first time that they have the opportunity to, but I'm removing night 1 restrictions as part of this PR so that amnesiac _can_ be matchmaker, wild child, etc. and that exchanging into matchmaker lets you act and choose another set of lovers.

longterm: I don't see the point of this. piper is covered from win stealer, and amnesiac-dullahan can still win because they'll get a much smaller list n3 than they would n1.

unsafe: this confuses me, and seems like it is already covered by neutral combined with some other tag.

tertiary: don't add tags unless they're going to actually be used right now (i.e. as part of this PR)

safe: Seer seeing villager vs wolf as the default makes far more sense than the reverse; we have more roles that are seen as one of those two categories than roles that are seen as the actual role. Plus, this tag does far more than impact seer -- it also has meaning for wolf mystic and the evilvillage game mode. Neatly enough, the roles that need special handling for all 3 of those things are exactly the same; it's not a case where one is a subset of the other or diverge a bit, it's literally the same roles in all 3 cases. There is no reason to split that into multiple tags.

secondary roles: we need to be able to specify actual role names in addition to categories. Otherwise, that syntax looks fine, and I like how it mostly follows the same format as role guide.

Vgr255:
Ah, I didn't realize mystic was using that category as well. I still think that seer should, by default, see the actual role that the person is, but if you tell me that it's better that way, I'll take your word for it! I agree with all the other points :)

PR #350, opened 2018-07-17
Rework how the !rolestats command parses its arguments

random-nick:
Now it tries completing the everything as a role first before attempting to complete it as a gamemode.

### DIFF
--- a/src/utilities.py
+++ b/src/utilities.py
@@ -376,7 +376,7 @@ def get_victim(cli, nick, victim, in_chan, self_in_list=False, bot_in_list=False
 
 class InvalidModeException(Exception): pass
 
-def complete_role(var, wrapper, role):
+def complete_role(var, role):
     if role not in var.ROLE_GUIDE.keys():
         special_keys = {"lover"}
         evt = Event("get_role_metadata", {})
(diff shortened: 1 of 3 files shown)


src/utilities.py
```diff
@@ -387,12 +387,10 @@ def complete_role(var, wrapper, role):
...
-            wrapper.reply(messages["no_such_role"].format(role))
-            return False
+            return []
         if len(matches) > 1:
-            wrapper.reply(messages["ambiguous_role"].format(", ".join(matches)))
-            return False
-        return matches[0]
-    return role
+            return matches
+        return matches
```
Vgr255: This whole block can simply be replaced by `return matches` right after the `complete_match` call.

src/wolfgame.py
```diff
@@ -5586,35 +5586,51 @@ def role_stats(var, wrapper, rest):
...
-    elif len(rest) == 1 or (rest[-1] == "all" and rest.pop()):
-        role = complete_role(var, wrapper, " ".join(rest))
-        if role:
-            wrapper.reply(db.get_role_stats(role))
-    else:
-        role = complete_role(var, wrapper, " ".join(rest[:-1]))
-        if not role:
+        return
+
+    role = complete_role(var, rest)
```
Vgr255: Naming nit: Since there may be multiple roles returned, name the variable `roles` instead. This helps reduce confusion for other people looking at the code. Are we doing something with the first letter of the role?

src/wolfgame.py
```diff
@@ -5586,35 +5586,51 @@ def role_stats(var, wrapper, rest):
...
-            if len(matches) > 1:
-                wrapper.pm(messages["ambiguous_mode"].format(rest[1], ", ".join(matches)))
-                return
-        wrapper.reply(db.get_role_stats(role, gamemode))
+
+    if len(params) == 1:
+        wrapper.pm(db.get_role_totals(gamemode))
+        return
+    
+    role = complete_role(var, " ".join(params[:-1]))
```
Vgr255: Same here for variable name.

PR #353, opened 2018-10-03
Split jester & fool

Vgr255:
I removed the hack with the `@winner` thing for fool, replacing it with a global tracking var for the lynched fool. Slight change in `transition_day` and `transition_night`, where instead of ensuring the phase is not day (or night), I now ensure the phase is either join or night/day. This is required to make fool work now that it's inside a listener.

### DIFF
--- a/src/roles/assassin.py
+++ b/src/roles/assassin.py
@@ -70,14 +70,12 @@ def on_transition_night_end(evt, var):
         random.shuffle(pl)
         pl.remove(ass)
 
-        if ass in get_all_players(("village drunk",)): # FIXME: Make into an event when village drunk is split
-            TARGETED[ass] = random.choice(pl)
-            PREV_ACTED.add(ass)
-            message = messages["drunken_assassin_notification"].format(TARGETED[ass])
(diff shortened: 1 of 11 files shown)


src/gamemodes.py
```diff
@@ -117,7 +117,7 @@ def set_default_totem_chances(self):
     # Here so any game mode can use it
     def lovers_chk_win(self, evt, var, rolemap, mainroles, lpl, lwolves, lrealwolves):
         winner = evt.data["winner"]
-        if winner is not None and winner.startswith("@"):
+        if winner in Win_Stealer:
```
skizzerz: This is wrong. Lovers not winning with fool is a special case due to how easy it is to set up that win condition with lovers. Lovers can (and should) win with other win stealing roles like monster, demoniac, piper, etc.

Also, the name of that variable is all sorts of wrong. Make it CamelCase without underscores.

Vgr255: I'll look into it later.

Also, which variable? Win_Stealer? It's been like that since I added globals for cats. I can change it but not as part of this PR.

PR #354, opened 2018-10-04
Add the absent status and split relevant code

Vgr255:
This simplifies a lot of code, and streamlines its handling.

### DIFF
--- a/src/status/absent.py
+++ b/src/status/absent.py
@@ -1,0 +1,55 @@
+from src.decorators import event_listener
+from src.containers import UserDict
+from src.functions import get_players
+from src.messages import messages
+from src import channels
+
+__all__ = ["add_absent"]
+
+ABSENT = UserDict() # type: UserDict[users.User, Tuple[str, bool]]
+
+def add_absent(var, target, key, *, private=True):
+    if target not in get_players():
+        return
+
+    ABSENT[target] = (key, private)
+
+    for votee, voters in list(var.VOTES.items()):
+        if target in voters:
+            voters.remove(target)
+            if not voters:
+                del var.VOTES[votee]
+            break
+
+    from src.wolfgame import chk_decision, chk_win
+    chk_decision()
+    chk_win()
+
+@event_listener("del_player")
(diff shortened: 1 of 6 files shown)


messages/en.json
```diff
@@ -671,7 +670,6 @@
     "succubus_target_success": "You are entrancing \u0002{0}\u0002 tonight.",
     "no_kill_succubus": "You discover that \u0002{0}\u0002 is a succubus and have retracted your kill as a result.",
     "no_target_succubus": "You discover that \u0002{0}\u0002 is a succubus and must now target someone else.",
-    "drunk_target": " In your drunken stupor, you have selected \u0002{0}\u0002 as your target.",
```
skizzerz: ?

src/status/absent.py
```diff
@@ -0,0 +1,55 @@
...
+from src.containers import UserDict
+from src.functions import get_players
+from src.messages import messages
+from src import channels
+
+__all__ = ["add_absent"]
+
+ABSENT = UserDict() # type: UserDict[users.User, Tuple[str, bool]]
+
+def add_absent(var, target, key, *, private=True):
```
skizzerz: For conformance to all other statuses, the key should be a "reason" (aka prefix) so that multiple messages could be sent out and customized using the specified reason.

Remove the private kwarg and make it always send a pm.

src/status/absent.py
```diff
@@ -0,0 +1,55 @@
...
+    for votee, voters in list(var.VOTES.items()):
+        if target in voters:
+            voters.remove(target)
+            if not voters:
+                del var.VOTES[votee]
+            break
+
+    from src.wolfgame import chk_decision, chk_win
+    chk_decision()
+    chk_win()
```
skizzerz: Status modules should never have side effects like phase transitions. Move these out of add_absent and handle them by the code which calls them if necessary (priest and gunner are the only two I can think of which would need to).

messages/en.json
Vgr255: Weird, I thought I'd removed that in my drunk branch. Either way it's not used (the full message is used instead)

PR #367, opened 2019-02-19
Master

Iciloo:
Add "[...] saw a job listing for "villager leader", and so joined the village because that is the best message ever


nyuszika7h:
Uhh, try again...

PR #372, opened 2019-05-16
Message refactor

skizzerz:


src/users.py
```diff
@@ -268,6 +268,11 @@ def __str__(self):
     def __repr__(self):
         return "{self.__class__.__name__}({self.nick!r}, {self.ident!r}, {self.host!r}, {self.realname!r}, {self.account!r}, {self.channels!r})".format(self=self)
 
+    def __format__(self, format_spec):
+        if format_spec == "@":
+            return "\u0002" + self.name + "\u0002"
```
Vgr255: Maybe `"\u0002{0}\u0002".format(self.name)` instead?

PR #392, opened 2019-09-28
Tweak wording for silver bullet messages

ammongit:
The tenses and wording was slightly awkward.


ishanyx:
this removes the distinction between an accidental headshot and a sharpshooter headshot - is that intentional?

skizzerz:
No, that distinction needs to exist

ammongit:
Alright, will adjust

nyuszika7h:
I think the way it is currently makes sense too, I mean the process of death can take a while, though not too long because the player is immediately removed from the game.

PR #395, opened 2019-10-15
Fix import in wolfgame.leave

mweinelt:
The pregame module is relative to the src package, so fix the import.


skizzerz:
Please make PRs based against the master branch, not stable.

mweinelt:
Oh well, that change is already on master. An here I thought I was running master. My bad.

PR #422, opened 2020-03-25
Add warnings for people who idle out night

### LINKED ISSUE #421: turncoat doesn't make it clear when you can change sides, or must pass
Turncoat can change sides only every other night, but the message is identical every night. Also, it doesn't tell you you can pass, or that you should pass if you don't want to change sides.

https://www.irccloud.com/pastebin/XvquMkaa/

skizzerz:
At most 1 warning point will be given per game per player. If they idle
out entirely, that will replace the night idle warning.

Certain roles/modes will make players immune to being warned for idling
out night. Right now, this is just time lord (once activated) and sleepy
(for players having a nightmare), as it is expected that more night
idles will happen due to those and that is part of expected gameplay.

Also fix up a couple of roles that were abusing chk_nightdone. It *must*
only store users who can act and have acted *that* night, not prior
nights.

Finally, shore up turncoat's messaging. Fixes #421


skizzerz:
Note: entirely untested as of now. Should be tested before merging.

jacob1:
Might also consider not applying the warning in case of exchange, or other role swapping happening in the middle of the night (which is rare but can happen if someone idles out for example)

messages/en.json
```diff
@@ -999,7 +999,8 @@
     "vg_against": "against {0!role:plural}",
     "vengeful_ghost_revealroles": "[b]dead vengeful ghost[/b]: {0:join}",
     "sleepy_revealroles": "[b]having nightmare[/b]: {0}",
-    "turncoat_notify": "You are {=turncoat!role:article} {=turncoat!role:bold}. You can change which team you're siding with every other night. Use \"{=side!command} {=villager!role:plural}\" or \"{=side!command} {=wolf!role:plural}\" to select your team.",
+    "turncoat_notify": "You are {=turncoat!role:article} {=turncoat!role:bold}. You can change which team you're siding tonight. Use \"{=side!command} {=villager!role:plural}\" or \"{=side!command} {=wolf!role:plural}\" to select your team. Use \"{=pass!command}\" if you do not wish to change teams tonight.",
```
Vgr255: which team you're siding with*

PR #433, opened 2020-04-18
Fix lynch stale state

soraxas:
There was an issue that I encounter earlier when testing locally, where the engine would be stated during the lynching state.

It happens when the game only had three people left, where one of them abstain and the other two choose to lynch each other.

i.e. For person A, B, C:
```
A -> B
B -> A
C (no lynch)
```

Since the needed majority votes is ` needed = avail // 2 + 1` = `2`, none of the condition is met. Force lynch is not set on the server so there was neither any people have majority votes to be lynched nor the abstains votes being the majority.

I think in this scenario if force lynch is not set, it should be seen as abstaining for the day.

### DIFF
--- a/src/votes.py
+++ b/src/votes.py
@@ -246,6 +246,9 @@ def chk_decision(var, *, timeout=False):
                         to_vote.extend(voting)
                 else:
                     abstaining = True
+            elif len(ABSTAINS | get_forced_abstains(var)) + len(VOTES) == avail:
+                # all people had voted and no conclusion can be made
+                abstaining = True
 
         if abstaining:
             for forced_abstainer in get_forced_abstains(var):


jacob1:
This is working as intended. There are two voting mechanisms: normal voting, and abstaining. They are separate, not meant to be directly related. So either a majority of voters need to vote someone out (2 in this case), or 50%+ of voters need to abstain (also 50% in this case). If two people vote each other, and one abstains, game shouldn't end. The stalemate needs to be resolved. Probably the person abstaining should not abstain, because in a 3p situation with one wolf, abstaining is always the wrong answer.

In nearly all cases, you can resolve stalemates like this. Try to convince whoever you can to vote along with you. The primary case where you can't is when everyone is on different teams (such as one wolf, one villager, and one fool). But in cases with one wolf and two villagers (or one wolf, one villager, and one crazed shaman), this stalemate should not exist.

soraxas:
Thanks for the explanation! The actual situation that I encountered was actually more like 8 people and in a state that nothing `>= 50%`, the 3p was just me testing it.

I understand how this implementation could encourage people to lynch more (and in fact, I realised even in this situation, the `timeout` for daytime would forcefully end the day with `sunset` message, but I had disabled the `TIMER` for my group so it became stale as no one want to change their vote; I just used `!fnight` to end the vote).

I understand if this doesn't fit the scope of the project. Now, an anonymous vote variant would be interesting to implement :) (coming from voting mechanism like avalon) as a more incomplete information game

jacob1:
Yeah when I played this IRL many years ago, we'd all go around in order and vote. If there wasn't a majority at the end, nobody would be voted out.

The balance I played with IRL is different than the balance in lykos. In lykos, anyone can vote at any time, and we wait until someone has been voted out. So that's why we intentionally don't end day in this situation here. We want someone to be voted out every day, unless the village agrees to !abstain.

If your players are still unable to agree on a vote, !fnight is acceptable.

Not sure if anonymous voting would make sense for lykos. You're free to try coding that in a custom bot like yours.

PR #434, opened 2020-04-18
Add settings to disable moderating

soraxas:
This is an excellent repo, and when I was hosting some friendly game with friends recently (with webcams) I found the auto bot moderating to be a bit annoying (as we all know each other and doesn't need the bot to do timers or idle moderating). I saw these settings exist for debug mode but not in normal mode.

These changes allow `botconfig.py` to customise the moderating behaviour of the bot.

### DIFF
--- a/src/settings.py
+++ b/src/settings.py
@@ -132,6 +132,12 @@ STATS_TYPE = "default" # default/accurate/team/disabled - what role information 
 START_VOTES_SCALE = 0.3
 START_VOTES_MAX = 4
 
+# Custom settings to disable auto bot moderating
+DISABLE_TIMERS = False
+DISABLE_TIME_LORD = False
+DISABLE_REAPER = False
+DISABLE_STASIS = False
+
 # Debug mode settings, whether or not timers and stasis should apply during debug mode
 DISABLE_DEBUG_MODE_TIMERS = True
 DISABLE_DEBUG_MODE_TIME_LORD = False
--- a/src/wolfgame.py
+++ b/src/wolfgame.py
@@ -126,7 +126,7 @@ var.DISCONNECTED = UserDict() # type: UserDict[User, Tuple[datetime, str]]
 
 var.RESTARTING = False
 
-if botconfig.DEBUG_MODE and var.DISABLE_DEBUG_MODE_TIMERS:
+if var.DISABLE_TIMERS or (botconfig.DEBUG_MODE and var.DISABLE_DEBUG_MODE_TIMERS):
     var.NIGHT_TIME_LIMIT = 0 # 120
     var.NIGHT_TIME_WARN = 0 # 90
     var.DAY_TIME_LIMIT = 0 # 720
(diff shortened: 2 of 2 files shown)


jacob1:
Hmm. Maybe we only need DISABLE_REAPER, and not the others? I can also see the argument for DISABLE_TIMERS though. You could also just extend DAY_TIME_WARN, DAY_TIME_LIMIT, SHORT_DAY_WARN, and SHORT_DAY_LIMIT in your botconfig.py instead of disabling them outright.

I definitely think DISABLE_TIME_LORD would be a silly option to have. It's meant to be a role that when accidentally lynched, makes the timers become really short, requiring quick action for all players deciding who to lynch / kill.

But anyway, I'm not specifically opposed to these settings. I do think more options for more scenarios would be cool. If you're using webcams and voice chat, the reaper is entirely useless to you, and you might want to extend the day / night cycle timers.

And I also like how you were running your games! Never heard of someone using lykos with webcams like that before, that's pretty cool.

skizzerz:
You can just directly set the appropriate things in botconfig.py, we don't need extra meta-settings that set other settings. You can copy basically anything in src/settings.py into botconfig.py to customize it.

For example, adding the following in botconfig.py disables the reaper and warnings system:
```
KILL_IDLE_TIME = 0
WARN_IDLE_TIME = 0
PM_WARN_IDLE_TIME = 0
PART_GRACE_TIME = 0
QUIT_GRACE_TIME = 0
ACC_GRACE_TIME = 0
LEAVE_PENALTY = 0
IDLE_PENALTY = 0
PART_PENALTY = 0
ACC_PENALTY = 0
NIGHT_IDLE_PENALTY = 0
```

I **do not recommend** disabling night/day timers, even if you're playing with webcams. Forcing people to make decisions is a good thing.

skizzerz:
Please see [our wiki page on Configuration](https://werewolf.chat/Configuration) for more information on how to configure the bot. The wiki page doesn't exhaustively list every setting you can actually modify in botconfig.py, but the note at the top does mention that basically everything in src/settings.py can be modified by adding it to botconfig.py. So, the bot is likely a fair bit more configurable than you originally thought :)

Please keep in mind that changing certain things greatly impacts game balance, namely anything that names a particular role. These settings are meant to be changed by game modes, not set as bot-level config.

The following are not settings (despite being in settings.py) and should never be changed: ALL_FLAGS, GRAVEYARD_LOCK, WARNING_LOCK, and all settings of the form RW_*

soraxas:
Hi @jacob1 @skizzerz  thank you for the timely review!

To explain myself, the reason of including `DISABLE_TIME_LORD` originally was simply because I want to be consistent with the other respective `DEBUG` options. Indeed you are correct, that what I really cared about is the `REAPER` and `TIMERS` only (`STASIS` isn't really needed if the other two are disabled).

While skizzerz you are correct of saying all those settings could achieve the same effect, I wouldn't say they are reluctant in the sense that: 
1. You can only track down that list by sniffing the code and knowing the value zero (or in fact `any value <= 0`) would have a special effect of disabling the `threading.Timer`, etc.;
2. No other unforeseen consequence (as one would assume timer should be positive?);
3. And importantly such a long list already demonstrate how unintuitive it is.

Having said that, I would go along with the decision and will continue to just use my own fork for that (along some other things). The PR is just a mere effort to contribute back for such an amazing project; and yes @jacob1 :) we are self-hosting the game because our group of boardgame gangs love playing the actual card game face-to-face but couldn't due to the current self-isolation circumstance. I've looked into a lot of other open-source werewolf projects, and this is the most support and features-rich (and support IRC!!) that I could found. So instead of playing together face-to-face, our boardgames night are now online and webcamed :)

skizzerz:
All of the various timer settings I listed above won't be changing in any meaningful way until the point where the config system is changing entirely and things will move from having a botconfig.py to having a botconfig.yml configuration-based system (should be before the end of the year).

I'd be interested to hear what other changes you decided you required for your fork. If you drop by #lykos on chat.freenode.net we can maybe talk about whether or not some of those things would fit better upstream.

PR #452, opened 2020-08-22
Revamped config system

skizzerz:
Move config into yaml and stop mixing bot configuration with game state together into a single object.


src/handler.py
```diff
@@ -214,142 +209,63 @@ def latency_pong(cli, server, target, ts):
         wrapper.reply(messages["latency"].format(lat))
         hook.unhook(300)
 
-def run_lagcheck(cli):
```
Vgr255, line 217: Is the lagcheck going to live on anywhere?

src/db/__init__.py
```diff
@@ -906,11 +905,11 @@ def _upgrade(oldversion):
                 # Update FKs to be deferrable, update collations to nocase where it makes sense,
                 # and clean up how fool wins are tracked (giving fools team wins instead of saving the winner's
                 # player id as a string). When nocasing players, this may cause some records to be merged.
-                with open(os.path.join(dn, "db", "upgrade2.sql"), "rt") as f:
```
Vgr255, line 909: Don't forget to move the files so these lines are correct

Vgr255 (commented):
First draft review. Mostly so I know what to do when I work on it soon

src/pregame.py
```diff
@@ -303,32 +315,45 @@ def start(var, wrapper, *, forced=False, restart=""):
...
+        elif wvs > (len(villagers) / 2):
+            wrapper.send(messages["too_many_wolves"])
+        else:
+            need_reset = False
+
+        if need_reset:
+            wrapper.send(messages["default_reset"])
+            stop_game(ingame_state, abort=True, log=False)
+            return
+
```
Vgr255, line 334: This was actually a conflict of adding it when it didn't exist in master. I don't know what the appropriate solution is, so I'm going to leave the modified PR version in for now.

PR #454, opened 2020-10-08
Don't show counts less than MIN_PLAYERS in !roles

nyuszika7h:
Roles defined for lower counts are automatically merged into the lowest player count that is allowed in the botconfig.


skizzerz:
Since you're touching this anyway, mind making it so that the roles displayed actually get passed through the messages system for translations?

nyuszika7h:
Yeah, I added that. Anything else, or can we merge this now?

PR #463, opened 2020-12-21
RFC: modify guardian_lose_with_guards win condition in guardian

bugmen0t:
Recent guardian game [~2010-12-20 23:00 UTC] had this rather surprising interaction:
The day started with 3 wolves alive, 2 GA, and 1 vil alive.
wolf gunner shoots and wounds one of GA, and gets lynched afterward.
On their lynch, guardian_lose_with_guards gets triggered.
Triggering wolf win condition on wolf lynch seems counterintuitive, so this patch seeks to alleviate that.

An alternate fix would be switching `lwolves == lguardians` to `lwolves >= lguardians` so wolves would get their win when GA was shot, not sure what was an intention here.

### DIFF
--- a/src/gamemodes/guardian.py
+++ b/src/gamemodes/guardian.py
@@ -49,7 +49,8 @@ class GuardianMode(GameMode):
 
     def chk_win(self, evt, var, rolemap, mainroles, lpl, lwolves, lrealwolves):
         lguardians = len(get_players(["guardian angel", "bodyguard"], mainroles=mainroles))
-
+        lpl_with_absent = len(get_players)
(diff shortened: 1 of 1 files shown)


bugmen0t:
Full game logs: https://0bin.net/paste/VqiMXVUu#foW3PCl2Srn2dNtB6dsGqDic5MbyTkrNA4X2xIxch7d

skizzerz:
The point of the absent status is that those people are counted as not in the village for all win condition checking. It should probably be `>=` instead of `==` though so it ended earlier.

PR #470, opened 2021-05-29
Update type annotations to the point where mypy is satisfied

iamgreaser:
NOTE: Testingwise I ended up running this and doing a `!join` then a `!quit`, although that was after the first commit and I missed one `from __future__ import annotations` line so I needed another commit. This does touch a lot of files and I'm not sure how you want this PR to be arranged.

If a `botconfig.py` is not provided then running mypy like this (available on pip) should pass:

```
python3 -m mypy .
```

This is good enough to get mypy 0.812 to typecheck things successfully in non-strict mode, although a custom mypy.ini was provided to ignore all failures in `src.messages.*`.

A large chunk of type comments were turned into Python 3.5/3.6 inline annotations.

`from __future__ import annotations` allows Python 3.7+ to not actually attempt to resolve type names when it sees said inline annotations.

Various imports are guarded by the `typing.TYPE_CHECKING` variable so not to introduce potential import loops into the code.

Several assertions were put in places to make mypy happy. In most cases, if the assertions were not there, the code would have crashed anyway.

`UserSet` and `UserList` were confirmed to not need to be generic. `UserDict` on the other hand seemed like the key or the value could be not a `User` and therefore both key and value needed to be generic.

Several cases of `users.User`, mostly those in annotations and type comments, were changed to be just `User`.

`src/settings.py` was for the purposes of this commit deemed to be something longing to be killed with fire and as a result I've largely avoided giving it proper type annotations for some things, especially where it gets extra stuff dumped into it for no reason from `src/wolfgame.py`.

A bug was fixed in `src/status/dying.py` as a result of needing to get the types checked. `None` is falsy, but it was meant to return `True` rather than do a no-value return (which returns `None`). I'm not sure if we want to split this one out or not. (The other case of a no-value return being changed was due to it actually needing to return `None`, so the change is behaviourally-identical.)

### DIFF
--- a/src/cats.py
+++ b/src/cats.py
@@ -20,6 +20,8 @@
 # Village Objective: If all of the players with this cat are dead, the village wins.
 #    Only main roles are considered for this.
 
+from __future__ import annotations
+
 from collections import defaultdict
 from typing import Dict
 import itertools
@@ -32,10 +34,10 @@ __all__ = [
     "Spy", "Intuitive", "Cursed", "Innocent", "Team_Switcher", "Wolf_Objective", "Village_Objective", "All"
 ]
 
-_dict_keys = type(dict().keys())
+_dict_keys = type(dict().keys())  # type: ignore
 
 # Mapping of category names to the categories themselves; populated in Category.__init__
-ROLE_CATS = {} # type: Dict[str, Category]
+ROLE_CATS: Dict[str, Category] = {}
 
 # the ordering in which we list roles (values should be categories, and roles are ordered within the categories in alphabetical order,
 # with exception that wolf is first in the wolf category and villager is last in the village category)
@@ -44,7 +46,7 @@ ROLE_ORDER = ["Wolf", "Wolfchat", "Wolfteam", "Village", "Hidden", "Win Stealer"
 
 FROZEN = False
 
-ROLES = {}
+ROLES: Dict[str, str] = {}
 
 def get(cat):
     if not FROZEN:
--- a/src/cats.py
+++ b/src/cats.py
@@ -46,7 +46,7 @@ ROLE_ORDER = ["Wolf", "Wolfchat", "Wolfteam", "Village", "Hidden", "Win Stealer"
 
 FROZEN = False
 
-ROLES: Dict[str, str] = {}
+ROLES = {}
 
 def get(cat):
     if not FROZEN:
--- a/src/context.py
+++ b/src/context.py
@@ -122,9 +122,7 @@ def context_types(*types):
 class IRCContext:
     """Base class for channels and users."""
 
-    _messages: Dict[str, List[IRCContext]] = defaultdict(list)
-
-    is_fake: ClassVar[bool] = False
+    _messages = defaultdict(list)
 
     def __init__(self, name, client):
         self.name = name
@@ -321,8 +319,7 @@ class IRCFeatures:
         modes = self._features.get("CHANMODES", [])
         while len(modes) < 4:
             modes.append("")
-        rA, rB, rC, rD = modes[:4]
-        return (rA, rB, rC, rD)
+        return tuple(modes[:4])
 
     @CHANMODES.setter
     def CHANMODES(self, value: str):
--- a/src/context.py
+++ b/src/context.py
@@ -1,8 +1,9 @@
 from __future__ import annotations
+
 import sys
 from collections import defaultdict, OrderedDict
 from operator import attrgetter
-from typing import Dict, Any, List, Set, Optional, Tuple
+from typing import Dict, Any, ClassVar, List, Set, Optional, Tuple
 
 import src.settings as var # FIXME
 from src.messages.message import Message
@@ -121,7 +122,9 @@ def context_types(*types):
 class IRCContext:
     """Base class for channels and users."""
 
-    _messages = defaultdict(list)
+    _messages: Dict[str, List[IRCContext]] = defaultdict(list)
+
+    is_fake: ClassVar[bool] = False
 
     def __init__(self, name, client):
         self.name = name
@@ -305,11 +308,11 @@ class IRCFeatures:
         self._features["CHANLIMIT"] = {}
         parts = value.split(",")
         for part in parts:
-            prefixes, limit = part.split(":")
-            if limit == "":
-                limit = None
+            prefixes, limit_str = part.split(":")
+            if limit_str == "":
+                limit: Optional[int] = None
             else:
-                limit = int(limit)
+                limit = int(limit_str)
             for prefix in prefixes:
                 self._features["CHANLIMIT"][prefix] = limit
 
@@ -318,7 +321,8 @@ class IRCFeatures:
         modes = self._features.get("CHANMODES", [])
         while len(modes) < 4:
             modes.append("")
-        return tuple(modes[:4])
+        rA, rB, rC, rD = modes[:4]
+        return (rA, rB, rC, rD)
 
     @CHANMODES.setter
     def CHANMODES(self, value: str):
@@ -372,6 +376,7 @@ class IRCFeatures:
 
     @EXTBAN.setter
     def EXTBAN(self, value: str):
+        prefix: Optional[str]
         prefix, types = value.split(",")
         if not prefix:
             prefix = None
--- a/src/debug/dict.py
+++ b/src/debug/dict.py
@@ -1,6 +1,6 @@
-from typing import Iterator, TypeVar, Union, Set, Dict, Mapping, Iterable
+from typing import Generic, Iterator, TypeVar, Union, Set, Dict, Mapping, Iterable
 import collections.abc
-import botconfig
+import botconfig  # type: ignore
 from src.debug.history import History
 
 __all__ = ["CheckedDict"]
@@ -10,7 +10,7 @@ VT = TypeVar("VT")
 KT_co = TypeVar("KT_co", covariant=True)
 VT_co = TypeVar("VT_co", covariant=True)
 
-class CheckedDict(collections.abc.MutableMapping):
+class CheckedDict(collections.abc.MutableMapping, Generic[KT, VT]):
     """ Dict container with additional features to aid in debugging.
 
     Common mutation methods are exposed to more easily set breakpoints,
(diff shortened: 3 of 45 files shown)


src/cats.py
```diff
@@ -44,7 +46,7 @@
 
 FROZEN = False
 
-ROLES = {}
+ROLES: Dict[str, str] = {}
```
skizzerz, line 49: This is wrong. You also put `Dict[str, str]` in a handful of other places where it's wrong. Actually check how the type is being used before throwing an arbitrary annotation on it. If you can't figure it out, leave it un-annotated.

src/context.py
```diff
@@ -121,7 +122,9 @@ def wrapper(cls):
 class IRCContext:
     """Base class for channels and users."""
 
-    _messages = defaultdict(list)
+    _messages: Dict[str, List[IRCContext]] = defaultdict(list)
+
+    is_fake: ClassVar[bool] = False
```
skizzerz, line 127: Why is this being added here?

src/context.py
```diff
@@ -318,7 +321,8 @@ def CHANMODES(self) -> Tuple[str, str, str, str]:
         modes = self._features.get("CHANMODES", [])
         while len(modes) < 4:
             modes.append("")
-        return tuple(modes[:4])
+        rA, rB, rC, rD = modes[:4]
+        return (rA, rB, rC, rD)
```
skizzerz, line 325: Why?

src/debug/dict.py
```diff
@@ -46,7 +46,7 @@ def __delitem__(self, k: KT) -> None:
         self._history.add("delitem", k)
         del self._dict[k]
 
-    def __getitem__(self, k: KT) -> VT_co:
+    def __getitem__(self, k: KT) -> VT:
```
skizzerz, line 49: Why this change? Do you know what covariance is? I'm nearly certain this was correct before, because we may be returning a type that is covariant in VT, not necessarily an instance of VT itself.

src/decorators.py
```diff
@@ -23,22 +25,22 @@
...
 
-_local = _local()
+_local = _local_cls()
 
 # This is a mapping of stringified tracebacks to (link, uuid) tuples
 # That way, we don't have to call in to the website everytime we have
 # another error.
 
-_tracebacks = {}
+_tracebacks: Dict[str, str] = {}
```
skizzerz, line 43: Completely wrong type

src/decorators.py
```diff
@@ -192,7 +194,7 @@ def __exit__(self, exc_type, exc_value, tb):
 
 class handle_error:
 
-    def __new__(cls, func=None, *, instance=None):
+    def __new__(cls, func: Optional[Callable[..., None]]=None, *, instance=None):
```
skizzerz, line 197: func shouldn't be optional here; it defaulting to None is wrong (should have no default)

src/decorators.py
```diff
@@ -307,6 +309,7 @@ def caller(self, var, wrapper: MessageDispatcher, message: str):
                 return # commands not allowed in alt channels
 
         if "" in self.commands:
+            assert self.func is not None
```
skizzerz, line 312: rather than asserting this isn't None everywhere, simply modify the typing of self.func to indicate that it's not Optional

src/handler.py
```diff
@@ -192,8 +194,10 @@ def parse_and_dispatch(var,
                 wrapper.pm(messages["no_force_admin"])
                 return
             if fn.chan:
+                assert channels.Main is not None
```
skizzerz, line 197: this is never None in the course of normal bot operation (with exception of early bootstrapping), so it should be typed as such rather than adding spurious asserts elsewhere. Ditto users.Bot

src/lineparse.py
```diff
@@ -59,7 +61,8 @@ def add_subparsers(self, **kwargs):
         self.allow_intermixed = False
         return super().add_subparsers(**kwargs)
 
-    def parse_args(self, args: Optional[Sequence[str]] = None, namespace: Optional[Namespace] = None) -> Namespace:
+    def parse_args(self, args: Optional[Sequence[str]] = None, namespace: Optional[Namespace] = None) -> Namespace:  # type: ignore
+        # parse_args has a difficult type signature to follow
```
skizzerz, line 65: Revert this

src/messages/formatter.py
```diff
@@ -103,7 +103,7 @@ def format_field(self, value, format_spec, *, flatten_lists=True):
             value = self._article(value, specs["article"])
             del specs["article"]
         if "!" in specs:
-            from botconfig import CMD_CHAR
+            from botconfig import CMD_CHAR  # type: ignore
```
skizzerz, line 106: this is a str

src/pregame.py
```diff
@@ -17,16 +20,19 @@
...
 
 WAIT_LOCK = threading.RLock()
 WAIT_TOKENS = 0
 WAIT_LAST = 0
 
-LAST_START = UserDict() # type: UserDict[users.User, List[datetime, int]]
-LAST_WAIT = UserDict() # type: UserDict[users.User, datetime]
-START_VOTES = UserSet() # type: UserSet[users.User]
-RESTART_TRIES = 0 # type: int
+LAST_START: UserDict[User, List[Union[datetime, int]]] = UserDict()  # FIXME: this should probably be a class containing two fields
```
skizzerz, line 32: don't add FIXME comments to existing code without consulting the main devs. We do actually care about and track these.

src/roles/bodyguard.py
```diff
@@ -1,22 +1,26 @@
...
 from src.messages import messages
 from src.status import try_misdirection, try_exchange, add_protection, add_dying
 from src.cats import Wolf
+from src.users import User
 
-GUARDED = UserDict() # type: UserDict[users.User, users.User]
+GUARDED: UserDict[User, User] = UserDict()
 PASSED = UserSet()
-DYING = set()
+DYING: Set[User] = set()
```
skizzerz, line 23: Should be a UserSet() and additionally cleared in the reset event. Not sure why this wasn't the case already

src/roles/matchmaker.py
```diff
@@ -13,9 +16,12 @@
 from src.status import try_misdirection, try_exchange, add_dying
 from src.cats import Win_Stealer
 
+if TYPE_CHECKING:
+    from src.users import User
+
 MATCHMAKERS = UserSet()
 ACTED = UserSet()
-LOVERS = UserDict()
+LOVERS: UserDict[User, User] = UserDict()
```
skizzerz, line 24: Wrong type

src/roles/warlock.py
```diff
@@ -13,10 +16,13 @@
...
 
+if TYPE_CHECKING:
+    from src.users import User
+
 register_wolf("warlock")
 
-CURSED = UserDict() # type: UserDict[users.User, users.User]
-PASSED = UserSet() # type: UserSet[users.Set]
+CURSED: UserDict[User, User] = UserDict()
+PASSED: UserSet = UserSet()
```
skizzerz, line 25: is the PASSED annotation necessary? Type is already obvious

src/settings.py
```diff
@@ -170,7 +173,7 @@
 
 DOCTOR_IMMUNIZATION_MULTIPLIER = 0.135 # ceil(num_players * multiplier) = number of immunizations
 
-GAME_MODES = {}
+GAME_MODES: Dict[str, Any] = {}
```
skizzerz, line 176: Wrong type

src/status/dying.py
```diff
@@ -67,7 +67,7 @@ def kill_players(var, *, end_game: bool = True) -> bool:
     with var.GRAVEYARD_LOCK: # FIXME
         if not var.GAME_ID or var.GAME_ID > t:
             #  either game ended, or a new game has started
-            return
+            return True  # CHECK BEFORE PULLING: This used to say just "return" which appears to have been an actual bug. --GM
```
skizzerz, line 70: remove comment

src/status/forcevote.py
```diff
@@ -16,30 +19,30 @@
 # FORCED_TARGETS is used to see which targets are being force-voted for.
 # If multiple calls to force_vote() force different targets, the union of all of those targets is taken.
 
-FORCED_COUNTS = UserDict() # type: UserDict[users.User, int]
-FORCED_TARGETS = UserDict() # type: UserDict[users.User, UserSet]
+FORCED_COUNTS: UserDict[User, int] = UserDict()
+FORCED_TARGETS: UserDict[User, UserSet] = UserDict()
 
-def _add_count(var, votee : users.User, amount : users.User) -> None:
+def _add_count(var, votee : User, amount : int) -> None:
```
skizzerz, line 25: bad annotation format, no space before colons (yes I know it was that way before, but you can fix it since you're already touching those lines)

src/status/silence.py
```diff
@@ -1,10 +1,14 @@
...
+from typing import TYPE_CHECKING
+
 from src.containers import UserSet
 from src.decorators import event_listener
 from src.messages import messages
 
 __all__ = ["add_silent", "is_silent"]
 
-SILENT = UserSet() # type: UserSet[users.User]
+SILENT: UserSet = UserSet()
```
skizzerz, line 11: Again, is this necessary?

src/users.py
```diff
@@ -10,16 +12,16 @@
...
 from src.match import Match
 
-import botconfig
+import botconfig  # type: ignore
 
 __all__ = ["Bot", "predicate", "get", "add", "users", "disconnected", "complete_match",
            "parse_rawnick", "parse_rawnick_as_dict", "User", "FakeUser", "BotUser"]
 
-Bot = None # bot instance
+Bot: Optional[BotUser] = None # bot instance
```
skizzerz, line 20: Will only be none during bootstrap; main code should treat this as non-Optional

src/wolfgame.py
```diff
@@ -72,62 +74,65 @@
...
 # Game Logic Begins:
 
-var.LAST_STATS = None
-var.LAST_ADMINS = None
-var.LAST_GSTATS = None
-var.LAST_PSTATS = None
-var.LAST_RSTATS = None
-var.LAST_TIME = None
-var.LAST_GOAT = UserDict() # type: UserDict[users.User, datetime]
+var.LAST_STATS = None  # type: ignore
```
skizzerz, line 86: omit all ignores. If mypy is throwing an error for these lines, suppress the individual error instead of opting out of the type system entirely. If you don't know the type, omit the annotation. Otherwise fill in the correct type.

skizzerz (changes requested):
Thanks for your effort, but a lot of this is quite sloppy. As-is this is unmergeable.

I've asked for clarification in a number of areas that I wasn't personally very sure about, please give a solid explanation for your changes there or revert them.

src/wolfgame.py
```diff
@@ -953,6 +958,7 @@ def fjoin(var, wrapper: MessageDispatcher, message: str):
 
     parts = re.split(" +", message)
     to_join = []
+    assert isinstance(wrapper.target, Channel)
```
Vgr255, line 961: Why is this necessary?

PR #492, opened 2021-11-08
Allow replies to non-players to go to channel in join phase

### LINKED ISSUE #483: Response to !stats and !time sent via notice even though I issued the commands in-channel
This seems to only happen during the night; during the day it works as expected.

wctaiwan:
Fixes #483

### DIFF
--- a/src/dispatcher.py
+++ b/src/dispatcher.py
@@ -44,7 +44,7 @@ class MessageDispatcher:
         if self.private:
             self.source.send(*messages, **kwargs)
         elif (self.target is channels.Main and self.game_state is not None and 
-                ((self.game_state.current_phase != "join" amd self.source not in get_players(self.game_state)) or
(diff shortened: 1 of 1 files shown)


src/dispatcher.py
```diff
@@ -44,7 +44,7 @@ def reply(self, *messages, prefix_nick=False, **kwargs):
         if self.private:
             self.source.send(*messages, **kwargs)
         elif (self.target is channels.Main and self.game_state is not None and 
-                (self.source not in get_players(self.game_state) or
+                ((self.game_state.current_phase != "join" and self.source not in get_players(self.game_state)) or
```
skizzerz, line 47: I'd just get rid of this branch entirely. I can't think of any legitimately useful purpose to have it, and the nightchat one will only ever be relevant if the channel modes are misconfigured.

skizzerz (changes requested):

skizzerz (approved):

PR #517, opened 2023-03-20
Clean up lover logic

skizzerz:
This gets rid of a lot of nested loops by making proper use of the adjacency list datastructure already in place to traverse through the lovers graph. A second graph containing forward edges only is maintained for revealroles and endgame display, so that the main lovers graph can be pruned as people die and therefore simplify a lot of other logic.

For the second graph, we map in order of lower accountname to higher accountname, since this is guaranteed to be stable throughout the course of a game.

### DIFF
--- a/src/roles/matchmaker.py
+++ b/src/roles/matchmaker.py
@@ -67,7 +67,7 @@ def get_all_lovers(var: GameState) -> list[set[User]]:
 
     return lovers
 
-def get_lovers(var: GameState, player: User, include_player: bool = False) -> set[User]:
+def get_lovers(var: GameState, player: User, *, include_player: bool = False) -> set[User]:
     """ Get all alive players this player is currently in love with.
 
     :param var: Game state
--- a/src/roles/matchmaker.py
+++ b/src/roles/matchmaker.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import random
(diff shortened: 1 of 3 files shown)


src/roles/matchmaker.py
```diff
@@ -33,39 +48,36 @@ def _set_lovers(target1, target2):
...
+            cur = queue.pop()
+            all_lovers.discard(cur)
+            visited.add(cur)
+            queue |= LOVERS[cur] - visited
+
+        lovers.append(visited)
 
     return lovers
 
+def get_lovers(player: User) -> set[User]:
```
Vgr255, line 68: Also should include `var`.

src/roles/matchmaker.py
```diff
@@ -33,39 +48,36 @@ def _set_lovers(target1, target2):
...
-            for lvr in LOVERS[lover]:
-                if lvr in pl:
-                    lovers[-1].add(lvr)
-
-    while set() in lovers:
-        lovers.remove(set())
+    all_lovers = set(LOVERS.keys())
+    while all_lovers:
+        cur = all_lovers.pop()
+        visited = {cur}
```
Vgr255, line 56: Should this call `get_lovers` for every lover? Could pass a `visited` kwarg so it communicates well; would make sure the logic is identical.

PR #518, opened 2023-07-13
Add locations

Vgr255:
This adds locations to lykos, allowing more flexibility for current and future roles.

Current task list to be completed before merging:

- [X] Has Village Square, Graveyard, and individual player's houses
- [X] Absent has been removed entirely; voting is now only possible if in the village square (and roles can set a "prison" reason for other roles, preventing them from leaving their current location until night)
- [ ] Insomniac has not been updated (need to figure out how it needs to behave)
- [ ] Mad Scientist has not been updated (should target adjacent houses if die at night, otherwise current behaviour)
- [X] Players will always be in exactly one location
- [X] There is a read-only property `GameState.locations`, though I don't think it's currently being used
- [X] Visiting roles (harlot, succubus) change their location to the house they visit
- [ ] Wolves (and possibly hunter/vigilante too) need to target a location, rather than a person
- [ ] Currently undefined behaviour for if someone does not have a house (might be out of scope)
- [ ] No testing has been done


skizzerz:
This has some good ideas but also quite a few weird ones that I would not be willing to keep as-is, and given two things I'm going to close this out and do it separately:

1. Non-location-related changes as part of the PR
2. You're mostly busy on non-location things now
3. I need this integrated into my work for new Halloween mode which heavily uses locations

PR #520, opened 2023-10-14
Add vampires

skizzerz:
Vampires forms a new third team (along with thralls for now, which have no abilities). The vampire team wins the same way wolves do: by overpowering the village. If both vampires and wolves are in the game, neither can win until the other team is eliminated.

Vampires are able to bite players to drain their blood. A player drained this way is absent and silent the following day and keeps the "drained" status for the rest of the game. Biting someone that is already drained kills them. Unlike wolves, each vampire chooses their own target to bite each night (they cannot double up, nor bite other vampires).

Vampires have access to the vampire chat, which is observable via the spectate and fspectate commands alongside wolfchat. The same settings that control wolfchat also control vampire chat.

Monsters are immune to vampires as well, and amnesiacs currently cannot become any vampire-team role (to prevent introducing a 3rd team into modes currently balanced for just wolves vs. villagers). The amnesiac thing may change in the future but this simplified initial implementation by not requiring a bunch of balance changes or win condition tweaks in modes with their own custom win conditions.

This has been lightly tested, but could probably use additional testing before merging.


skizzerz:
Since I forgot to do this, notes to myself for later:

- Seer and oracle both see vampires as villager. Similar to piper, the list of drained victims can be used to narrow down vampires instead.
- Vampires have enough self-control to refrain from biting a visiting harlot
- Vigilantes can safely kill vampires (but not thralls)
- Augur sees vampires and thralls as red
- Mystic includes vampires in the evil villager count (which will likely require a new role category)
- Wolf mystic see vampires as their own category
- Detective identifies vampires normally
- Investigator treats vampire team as its own team

messages/en.json
```diff
@@ -901,6 +906,15 @@
     "charm_multiple_success": "You have charmed {0:@} and {1:@}.",
     "another_piper_charmed_multiple": "Another piper has charmed {0:@} and {1:@}!",
     "another_piper_charmed": "Another piper has charmed {0:@}!",
+    "no_target_vampire": "You may not bite another vampire.",
+    "already_bitten_tonight": "{0:@} is already being bitten by another vampire tonight; you must select a different target.",
+    "vampire_bite": "You have chosen to bite {0:@} tonight.",
+    "vampire_bite_vampchat": "{0:@} has chosen to bite {1:@} tonight.",
+    "retracted_bite": "You have retracted your bite.",
+    "retracted_bite_vampchat": "{0:@} has retracted your bite.",
```
MishaCatskill, line 914: s/your/their/

PR #531, opened 2024-11-30
Prevent a turned vigilante from killing existing vampires

meldra:


skizzerz:
Closing since this will be addressed in a different manner

PR #542, opened 2025-06-30
Add day/night count in sunrise/sunset messages

bojidar-bg:
I have not tested this code change locally, but it \*should\* work if my Python is right?
(I'll get around to testing it in a few days if you haven't merged it first)

Anyway, look at me submitting a PR as requested (:o


skizzerz (approved):
This doesn't add any counters for day 1 when we start with day (e.g. mudkip), but that doesn't seem important enough to block merging.

